PR #8, opened 2014-11-25
Port to Python 3 and fix for directly flashing .elf images

jaseg:
...doing the conversion of the lower segments to the bootloader format on the fly.


tprochazka:
Why this pull request was not merged?

jaseg:
I suppose it pretty radically breaks with the old tool's semantics that are replicated over the entire "maker" blogosphere, and there are people who don't like that. Feel free to use my fork (called flashi for disambiguation). I did not put a lot maintenance into it during the last few months since I did not work on the esp8266. If you are missing any functionality, feel free to send a pull request.

projectgus:
I'm going to close this PR as it seems like the two tools have sufficiently diverged for now. Flashing directly from ELF without an interim step still sounds like a useful addition, but I think it will need substantial reworking against the updated esptool.py - and it seems like you've moved on from this now anyhow. Sorry!

PR #13, opened 2014-12-30
Added procedure reading MAC STA and MAC AP address

CHERTS:
1. Added procedure reading MAC STA and MAC AP address
2. Fixed algoritm go into bootloader mode


projectgus:
I'm going to close this as esptool has changed a lot since it was opened. It'd still be nice to dump WiFi AP & STA MAC addresses, but I think it will need to be written again against the new code structure.

PR #20, opened 2015-02-06
Fixed esptool for modern operating systems (like Arch)

Manawyrm:


themadinventor:
Been there, reverted that (285ec621ff94c47d1cffa241a09902f35a9c0baf). It'll always break on some platforms. There is, as far as I can tell, no completely platform independent way of solving this in the shebang line. There are Arch specific fixes available though (https://wiki.archlinux.org/index.php/python).

I'll try to implement it inside the script when I have time. PRs are of course welcome, as are convincing arguments that I'm wrong :)

projectgus:
I've tried to find a definitive summary of "what you should put in the shebang", but I can't. which surprised me because Python is normally pretty good at definitive summaries that say "just do it this way"!

From what I can tell:

* Python docs recommend using [`/usr/bin/env python` with 2.7 on Unix](https://docs.python.org/2.7/using/unix.html#miscellaneous), and [`/usr/bin/env python3` with 3.x on unix](https://docs.python.org/3.4/using/unix.html#miscellaneous).

* `/usr/bin/env python2.7` may break bash under some circumstances, I think.

* Windows needs the Python launcher for Windows ([PEP-397](https://www.python.org/dev/peps/pep-0397/)), but once it has that then any of the above will work.

* Aside from Arch, no other OSes have dared to change default `python` from 2.x. (I could be wrong about that.)

So, the solution offered here _might_ work on all platforms (I can't find any reason that python2 won't work, at least.)

Another, better but harder, solution might be to support both Python 2.7 and 3.x. That way it won't matter which version people have installed or set as default. On the downside, there'd be a new scope for bugs that only appear on one version. #8 looks like it did some of this work, but only to support Python 3.x not 2.7 & 3.x. I might put some time into investigating how messy this option would be...

projectgus:
Oops, I just found the discussion of this at #21 also. Comment there [explains that 'python2' breaks on OS X 10.9](https://github.com/themadinventor/esptool/issues/21#issuecomment-73390067).

I'm going to close this in favour of the discussion in #21 for now, as it does seem like there's no easy fix. Sorry!

pfalcon:
>  I might put some time into investigating how messy this option would be...

Please start with converting the source to use standard Python 4-space indent instead of tabs.

projectgus:
> Please start with converting the source to use standard Python 4-space indent instead of tabs.

I don't think there are any tabs in the esptool.py source...?

pfalcon:
Oops, sorry, I believe at one time there were, and in another @themadinventor's project, ida-xtensa, they're still there. False alarm, anyway.

themadinventor:
@pfalcon Sorry about that, fixed in themadinventor/ida-xtensa@345236bac33f193a6ae67328e49c828af8d80eb9.

pfalcon:
@themadinventor : Thanks!

PR #34, opened 2015-05-01
More auto-reset improvements

projectgus:
These changes are mostly based on a discussion I had with @markusgritsch regarding 88baa738a5b7ee5 (PR #30). Lots of discussion in that commit thread.

The first commit here should improve the reliability of auto-resets (using either the "standard" or the NodeMCU reset scheme).

The second commit helps avoid something that is probably unique to Linux, where the CH340 doesn't always switch baud rates like it should - resulting in random slow uploads at times. This shouldn't have any side effects on other OSes.

Thanks for all your work on esptool. :)

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -150,6 +150,10 @@ class ESPROM:
                     return
                 except:
                     time.sleep(0.05)
+            # this is a workaround for the CH340 serial driver on current versions of Linux,
+            # which seems to sometimes set the serial port up with wrong parameters
+            self._port.close()
+            self._port.open()
         raise Exception('Failed to connect')
 
     """ Read memory address in target """
--- a/esptool.py
+++ b/esptool.py
@@ -128,26 +128,28 @@ class ESPROM:
     def connect(self):
         print 'Connecting...'
 
-        # RTS = either CH_PD or nRESET (both active low = chip in reset)
-        # DTR = GPIO0 (active low = boot to flasher)
-        self._port.setDTR(False)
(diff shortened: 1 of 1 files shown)


themadinventor:
Great work @projectgus, thanks!
Seems like my reference hardware setup is bad; it is too forgiving, I usually can't reproduce the connectivity problems ;)
I feel a little bad about not being able to test all module/adapter/os permutations atm, I'm thinking about setting up some kind of multi-platform CI. It would require a few boxes and modules though.

projectgus:
It'd be hard to cover everything I think. The other factor is host OS, especially for the CH340 adapters where the drivers are a bit flaky and probably different platform to platform (especially on Linux I think as they're developed from looking at the behaviour of the Windows drivers!)

I think you're doing a solid job, there are worse problems than having to rerun the flashing tool every now and then. :+1:

projectgus:
(Oops, I mean "worse potential problems" - didn't mean to suggest there _are_ worse problems currently in esptool - it works great!!) :)

markusgritsch:
@themadinventor
> Seems like my reference hardware setup is bad; it is too forgiving,
> I usually can't reproduce the connectivity problems ;)

For me at 115200 baud it also works reliably every first time. However, I prefer to upload at 500000 baud, which dramatically shortenes the waiting time, and at those high baud rates, I failes sometimes. Therefore the second outer reset loop from @projectgus is very welcomed :)

PR #37, opened 2015-05-12
Make elf2image command thread-safe

projectgus:
Hi,

Just another minor tweak from me. I run into this when running parallel builds of an esp firmware (the two elf2image commands run simultaneously).

Cheers,

Angus

***

Allows two elf2image commands to be run in parallel (ie parallal
make process).


themadinventor:
oh, good catch! thanks!

PR #46, opened 2015-06-01
Better fix for CH341 slow baud rate bug (closes #44)

### LINKED ISSUE #44: Programming speed is terribly slow sometimes
I've been bitten by this ever since 1st module I tried. For some modules, programming speed is *very* slow. Example below (using a patch to be submitted) shows that actual UART speed used by bootloader is likely 9600. The issues definitely depends on USB-serial adapter used. But what's most annoying is "persistent sticky" nature of the problem. For example, today I just unpacked a new board with CH341 adapter builtin, and first programming went fast, but all subsequent show speed as below. Google is pretty silent on this specific issue, so posting this, as this is more or less appropriate place for it hopefully. Please don't close this prematurely, let's collect stats and maybe find some pattern.

~~~~
Connecting...
Erasing flash...
Writing at 0x0000d400... (100 %) 
Written 55296 bytes in 60.84 seconds (7270.84 bit/s)...
Erasing flash...
Writing at 0x0004f000... (100 %) 
Written 259072 bytes in 287.08 seconds (7219.56 bit/s)...

Leaving...
0.12user 0.04system 5:50.45elapsed 0%CPU (0avgtext+0avgdata 9224maxresident)k
~~~~

projectgus:
Based on information in #44

Looks like if the kernel thinks the serial port is already set to
[baudrate], it won't set the baud rate in set_termios. Due to the
pre-4.0 bug of always chosing 9600 baud on open, it's possible for the
device to be set to 9600 but the kernel thinks it's set much faster.

Not sure what I was thinking with the other fix in de97c54a0. It did
seem a bit more reliable but possibly this was because when testing I
kept changing the baud rate(!)

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -64,7 +64,12 @@ class ESPROM:
             "\x04\x00\x0b\xcc\x56\xec\xfd\x06\xff\xff\x00\x00"
 
     def __init__(self, port = 0, baud = ESP_ROM_BAUD):
-        self._port = serial.Serial(port, baud)
+        self._port = serial.Serial(port)
+        # setting baud rate in a separate step is a workaround for
+        # CH341 driver on some Linux versions (this opens at 9600 then
+        # sets), shouldn't matter for other platforms/drivers. See
+        # https://github.com/themadinventor/esptool/issues/44#issuecomment-107094446
+        self._port.baudrate = baud
 
     """ Read bytes from the serial port while performing SLIP unescaping """
     def read(self, length = 1):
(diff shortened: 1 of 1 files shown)


pfalcon:
+1, this patch is tested well by me.

@themadinventor: ping.

pfalcon:
ping

pfalcon:
This patch is now available in https://github.com/pfalcon/esptool/tree/esp-open-sdk , to be used by https://github.com/pfalcon/esp-open-sdk

PR #48, opened 2015-06-10
Patch by Viktor aka pvvx

CHERTS:
- Fixed wrong size erase region for recording.
- Fixed creation of superfluous segments for the loader, if the no data in segments rodata and etc.
Patch by Viktor aka pvvx (https://github.com/pvvx)

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -189,12 +189,34 @@ class ESPROM:
             raise Exception('Failed to leave RAM download mode')
 
     """ Start downloading to Flash (performs an erase) """
-    def flash_begin(self, size, offset):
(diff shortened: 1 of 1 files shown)


pfalcon:
Squash these 2 commits?

projectgus:
I think the changes here are covered by #53 and 96e3386.

The only functionality that is unmerged from here is ignoring zero-length sections in firmware image files. If the image file has zero-length sections in it already, commands like `image_info` should display that. OTOH if generating a firmware image from an ELF file, it makes sense not to bother adding zero-length sections to the new image.

PR #49, opened 2015-06-15
elf2image: Read start address from ELF header instead of call_user_start

### LINKED ISSUE #24: elf2image: Use map file instead of xtensa-lx106-elf-nm

projectgus:
Allows custom or stripped entry point symbol name. Alternative approach to #47, should close #24.

Also includes a refactor of common code for running SDK tools.

This adds a dependency on toolchain -objdump but any setup where -nm and -objcopy are installed should have -objdump.

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -329,6 +329,25 @@ class ESPFirmwareImage:
         f.write(struct.pack('B', checksum))
 
 
+def run_tool(toolname, *args):
+    """ Execute a toolchain tool (either from xtensa toolchain or gcc binutils,
+        return a list of stdout lines of output"""
+    if os.getenv('XTENSA_CORE')=='lx106':
+        tool = "xt-%s" % toolname
+    else:
+        tool = "xtensa-lx106-elf-%s" % toolname
+    try:
+        proc = subprocess.Popen([tool]+list(args), stdout=subprocess.PIPE)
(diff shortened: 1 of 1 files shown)


projectgus:
I'm going to close this PR as I think it's more or less covered by a0c18eac57 which just got merged.

This version did some refactoring and used objdump instead of readelf, but they're basically the same thing.

Thanks @themadinventor for continuing to improve esptool.py. :)

PS You can probably close #24 now.

Issue #33, opened 2015-04-29
Add workaround for a bug in SPIEraseArea


igrr:
```SPIEraseArea``` function in the esp8266 ROM has a bug which causes extra area to be erased.
If the address range to be erased crosses the block boundary (block is 16 sectors), then **extra** ```head_sector_count``` sectors are erased.
If the address range doesn't cross the block boundary, then **extra** ```total_sector_count``` sectors are erased.
Where ```head_sector_count``` is the number of sectors from the start of erased area to the block boundary, and ```total_sector_count``` is the total number of sectors to be erased.

Espressif added the workaround to their version of flash tools. The trick is to modify the erased size:
```
    """ Start downloading to Flash (performs an erase) """
    def flash_begin(self, _size, offset):
        old_tmo = self._port.timeout
        self._port.timeout = 10

        area_len = int(_size)
        sector_no = offset/4096;
        sector_num_per_block = 16;
        #total_sector_num = (0== (area_len%4096))? area_len/4096 :  1+(area_len/4096);
        if 0== (area_len%4096):
            total_sector_num = area_len/4096
        else:
            total_sector_num = 1+(area_len/4096)
        #check if erase area reach over block boundary
        head_sector_num = sector_num_per_block - (sector_no%sector_num_per_block);
        #head_sector_num = (head_sector_num>=total_sector_num)? total_sector_num : head_sector_num;
        if head_sector_num>=total_sector_num :
            head_sector_num = total_sector_num
        else:
            head_sector_num = head_sector_num

        if (total_sector_num - 2 * head_sector_num)> 0:
            size = (total_sector_num-head_sector_num)*4096
            print "head: ",head_sector_num,";total:",total_sector_num
            print "erase size : ",size
        else:
            size = int( math.ceil( total_sector_num/2.0) * 4096 )
            print "head:",head_sector_num,";total:",total_sector_num
            print "erase size :",size


        if self.command(ESPROM.ESP_FLASH_BEGIN,
                struct.pack('<IIII', size, 0x200, 0x400, offset))[1] != "\0\0":
            raise Exception('Failed to enter Flash download mode')
        self._port.timeout = old_tmo
```

Perhaps after some cleanup this fix should be applied here as well.
This issue causes the first 0x1000 to be erased in some cases when writing near the end of flash. In turn, this leads to a corrupted flash image when writing two binaries (0x00000 and 0x40000) separately.

scargill:
[...]

PR #51, opened 2015-06-26
flash erase fix and serial terminal

jkent:
I rewrote and tested the code in issue #33, and also included a simple serial terminal I wrote for another project.  It is a very convenient way to debug since you don't have to worry about the RTS line changing on you.  If you want me to make it GPLv2+ I can, but I'd rather it be dual licensed.


kanflo:
+1, please merge
"workaround for bug in SPIEraseArea" is needed when firmware/0x40000.bin grows larger than 0x30000 bytes.

jkent:
I'm closing this to reopen as two separate pull requests, shortly after simplification of the fix.

PR #52, opened 2015-07-03
Workaround for DIO write protect issue, automatic entrypoint extraction

jmattsson:
I originally made these changes for the NodeMCU project. Most notably it works around the issue where the flash ends up in write-protect mode when DIO is used, and as such fixes https://github.com/themadinventor/esptool/issues/40. The hows and whys are already referenced within that thread by Marcel.

The second change is to make esptool extract the actual entry point from the ELF file rather than assume a hardcoded call_user_start.

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -263,6 +263,15 @@ class ESPROM:
 
         return data
 
+    """ Abuse the loader protocol to force flash to be left in write mode """
+    def flash_unlock_dio(self):
+        # Enable flash write mode
+        self.flash_begin(0, 0)
+        # Reset the chip rather than call flash_finish(), which would have
+        # write protected the chip again (why oh why does it do that?!)
+        self.mem_begin(0,0,0,0x40100000)
+        self.mem_finish(0x40000080)
+
     """ Perform a chip erase of SPI flash """
     def flash_erase(self):
         # Trick ROM to initialize SFlash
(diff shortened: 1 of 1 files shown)


omerk:
@jmattsson Thanks for this fix, I can verify that it solves the filesystem issues I've been having on the new NodeMCU boards.

@themadinventor esptool, without this fix, is currently broken for all NodeMCU R2 board users. Please consider merging this PR.

themadinventor:
Thanks! I've been inexcusably absent for a while now...

creationix:
Awesome work everyone!

thehose:
Hey folks,
I have a NodeMCU V2 board (from China) which came pre-loaded with Lua. I actually want to use Micropython, so I setup a build environment (OSX), built and loaded uPy on the board and while I get a prompt, it's really disabled. I get a memory check error.
screen shot 2015-08-24 at 1 57 58 pm
After a lot of google, I came across this issue with esptool.py and thought I'd found the problem. Got the latest version of the tool and changed the defaults for -fm/DIO and -fs/32m in the tool. 
Same results.. I thought maybe it's a uPy issue, so to test I tried to load the latest Lua 0.9.6 using the tool as ( esptool.py --port=/dev/tty.SLAB_USBtoUART write_flash 0 ~/Desktop/nodemcu_integer_0.9.6-dev_20150704.bin -fs 32m -fm dio -ff 40m) and same type of issue..
screen shot 2015-08-24 at 2 06 47 pm

Any advise would be appreciated.....

jmattsson:
The `MEM CHECK FAIL!!!` is actually completely(?) harmless, though it doesn't sound like it. During boot-up the system looks for various saved values (calibration etc) in RTC system memory. On a cold-boot (either power-on or releasing CH_PD line) the RTC memory is uninitialised, and this message gets printed. On a reboot you would not get it.

thehose:
I have a Nodemcu V2 development board and this only happens when I depress the "RST" button, which I have to do when I upload any firmware. Is this what you refer to as a reboot ?

PR #53, opened 2015-07-08
workaround for bug in SPIEraseArea

jkent:
This is a cleaner rewrite of the code in bug #33 by iggr.  Also added a fake flash_begin so flash_finish returns sucessfully.  See also #48 and #51.


jkent:
I'm curious if the behavior for DIO from #52 should actually be used as the default?

gelraen:
FYI, I did some analysis on how and why this fix works: https://cesanta.com/blog/esp8266-flash-erasing-issue.shtml

flyashi:
After debugging for hours, this solved my issue. Please accept this pull request and save future me's the hassle!

jkent:
@themadinventor *poke*

PR #54, opened 2015-07-08
added a terminal feature to run command

jkent:
A simple serial terminal I wrote for another project. It is a very convenient way to debug since you don't have to worry about the RTS line changing on you. If you want me to make it GPLv2+ I can, but I'd rather it be dual licensed.  (Split from #51)


projectgus:
Hey,

Sorry for ignoring this for so long. Turns out I just found out (yesterday) that pyserial packages a quite flexible serial program called miniterm.py, so everyone with esptool.py already has a terminal program!

I've linked that from the README and I think that solves the immediate need for this, so I'm going to close the issue. Sorry it dragged on for so long though.

Angus

jkent:
Neat, I didn't know about it either.  There is one problem it doesn't solve though:  getting boot messages immediately after your flashed firmware runs.  Maybe miniterm.py can be imported to do this?

projectgus:
I generally have good luck chaining esptool.py and a serial program together on the command line - I've found the OS layer and the USB/serial layer cache enough data that you don't lose anything in the short gap between esptool.py exiting and the next program starting.

If you do the same, do you lose data?

jkent:
I haven't tried chaining to be honest.  I'll give it a shot.  Thanks!

PR #55, opened 2015-07-14
Change shebang line to python2

gh2o:
Python 3 is default on some systems, so this avoids that ambiguity.


projectgus:
Hi gh2o,

Thanks for this. I think this is the same issue as #20. There is some discussion in there about the possibilities, would appreciate your input there if you have any other thoughts.

PR #61, opened 2015-09-20
Manage possible delay between write and read operation to avoid serialPort exception

sguilly:


projectgus:
Hi,

Thanks for sending this. It's hard to test with all combinations of OSes and adapters, so seeing fixes for other combinations is super helpful. :)

pySerial has a built-in timeout feature that I think we can leverage to get the same result you're seeing here. I have a few questions to help me understand the problem, though:

* I saw you said you're on Linux, what kernel version are you on?
* Which serial adapter?
* What Python and pySerial version?
* Can you please post the output from a failing operation, including the command line and the full trace of the serialPortException?

The other thing that surprised me in the patch is the very high timeouts - up to 2.25 seconds for one operation. Did you really find that waiting this long was necessary? Having a high timeout means that if an operation really does fail (bad serial connection, for example), the software won't respond quickly with an error.

Thanks. :)

Angus

sguilly:
Previously, the original file works fine but after an (OS) update, i had this issue.

Nom : Linux sguilly-Z97X 3.13.0-64-generic #104-Ubuntu SMP Wed Sep 9 12:36:12 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux

Python 2.7.6

The serial adapter is a chinese USB/RS232 model.

I will try to take time later to report serialPortException.

BR/

Stéphane

projectgus:
Thanks for the details, Stéphane. Much appreciated. :)

> The serial adapter is a chinese USB/RS232 model.

Do you know exactly what type? If it's unclear then you can find out by looking at the output from `dmesg | tail` just after you plug in the adapter.

How are you wiring the reset & GPIO0 lines on the ESP8266?

Cheers,


Angus

projectgus:
There's been some interesting PR just merged (#66) that fixes some bugs with synchronisation on reset. Can you try with the latest 'master' and see if the problem is fixed without these commits?

projectgus:
I'm going to close this for now, but if you're still seeing the problem with the latest esptool.py then please re-open it and we'll work through the issue.

Issue #21, opened 2015-02-07
#!/usr/bin/env python vs python2 - aka PEP 394 The "python" Command on Unix-Like Systems


gdamjan:
https://www.python.org/dev/peps/pep-0394/

It suggests to use the supported python version in the shebang, in this case it should be
`#!/usr/bin/env python2`

themadinventor:
The shebang line stuff is a clusterfsck. `/usr/bin/env python2` does not work out of the box on f.ex. Mac OS 10.9 (at least on my system). Specifying `python2` is of course correct, but it'll always break on some systems, and still, citing PEP394:
> for the time being, all distributions should ensure that python refers to the same target as python2

I guess the best solution is to make esptool compatible with both Python 2 and 3, keeping it `python`, and if not possible, switch to `python3` and force users to install it.

dpgeorge:
> I guess the best solution is to make esptool compatible with both Python 2 and 3

+1

gdamjan:
The other approach is to make esptool a proper package/module and register an `console_scripts` entry-point [1]. This way, when setup.py install it, it will create a proper script with the correct shebang for the installed python. 

[1] http://www.scotttorborg.com/python-packaging/command-line-scripts.html#the-console-scripts-entry-point

jiegec:
Or consider using #!/use/bin/env python2.7 ? Tested on Mac and Arch Linux

jiegec:
I also agree with @gdamjan

projectgus:
I've pushed an experimental branch that runs on Python 2 and 3.
https://github.com/themadinventor/esptool/compare/python2_and_3

I'm not 100% convinced this isn't making a maintenance trap for down the road, but the differences weren't too terrible. I haven't tested all operations, some things probably still don't work.

Any suggestions or testing is welcome, though.

dbrgn:
Python 2/3 support would be great :)

projectgus:
I recently became an Arch user myself, so I wrote up some tips/workarounds here:
https://github.com/themadinventor/esptool/wiki/Arch-Linux-Tips

(I'm using pip 'development mode' myself now.)

jrobeson:
@projectgus : I took a look at the branch.  I think it should start a little smaller.  Can we adapt the coding standards (defacto or otherwise) so that all byte literals are marked up with `b'`?  That'd go a long way to making 2/3 work and stay working.

projectgus:
[...]

PR #62, opened 2015-09-23
Change shebang in order to run with correct python version

waaaaargh:
Some distribution ship with python3 as their default python implementation, which leads to esptool being run by python3, with which it is of course not compatible.

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 #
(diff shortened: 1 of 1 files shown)


projectgus:
Unfortunately there's no easy solution to this that works on all platforms, apart from having esptool support both Python 2 & 3.

Please have a look through the discussion in #21. If you have a chance to test the python2_and_3 branch that's mentioned there, I'd really appreciate any feedback.

Thanks!

PR #76, opened 2015-12-29
Make flash_size option for write_flash case insensitive

gregersrygg:
So it's possible to map flash_size directly from Arduino´s `platform.txt` file (which prints out an upper case M):

    --flash_size {build.flash_size}


projectgus:
Hey,

Great addition. It only failed to build because pep8 (the Python style guide) requires `type=str.lower` (no spaces around the equals).

Could you please amend the commit and force push? The Pull Request will update automatically.


Angus

gregersrygg:
Whoops, I didn't see that. Fixed :)

projectgus:
Thanks! :)

PR #78, opened 2016-01-05
Added verify_flash option

newinnovations:
I have added the option to verify a portion of the flash against the image on disk.

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -583,6 +583,13 @@ def main():
     parser_read_flash.add_argument('size', help='Size of region to dump', type=arg_auto_int)
     parser_read_flash.add_argument('filename', help='Name of binary dump')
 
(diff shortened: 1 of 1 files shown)


projectgus:
Thanks for this, very useful feature!

A couple of minor comments:
* Could you please squash these commits in your branch and re-push (force push) as a single commit?
* Is it possibe please to print out the address of the first different byte, to help people diagnose problems as being invalid flash as opposed to just bad serial comms or something else?

newinnovations:
I have squashed the commits and added some more information when verify fails: number of differences and first failed address. Furthermore there is now an option to dump all failed locations: --diff=yes. Default is no.

projectgus:
Thanks for that! Sorry for the delay in updating.

Have merged, plus some additions and refactoring to add a couple of other options I had mind.

PR #83, opened 2016-02-09
Fixed read_mac for recent ESP modules.

jmattsson:
Newer ESP modules use a different OUI (5c:cf:7f) which can be accessed from
the MAC3 field in the OTP. When this field is non-zero, it must be used
to form the complete MAC address.


projectgus:
Thanks for the patch. Interesting situation, I wonder if that means Espressif have sold their first 32 million chips, exhausting the other two OUIs.

PR #85, opened 2016-02-20
Add command to read Chip ID

urmilparikh:
Reference: http://esp8266-re.foogod.com/wiki/System_get_chip_id_(IoT_RTOS_SDK_0.9.9)

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -274,6 +274,12 @@ class ESPROM:
             raise FatalError("Unknown OUI")
         return oui + ((mac1 >> 8) & 0xff, mac1 & 0xff, (mac0 >> 24) & 0xff)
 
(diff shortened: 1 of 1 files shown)


projectgus:
Merged as 10a3c9c (I tweaked the commit to add a link to the esp8266-re wiki.) github hasn't picked up the relationship somehow.

Thanks for this! Interestingly, when I was testing I noticed the top 8 bits are zero on a lot of my chips, and (as we know) the bottom 24 match the MAC. That's both using esptool.py and the SDK system_get_chip_id() function. Do you see that as well?

urmilparikh:
Thanks for merging.

Yes, I have noticed that top 8 bits are always zero in all ESPs that I have come across.
But that may change any day!

PR #80, opened 2016-01-28
Hard reset on disconnect

fdivitto:
argfile doesn't open as binary file without 'rb' and flashing fails (at least using Python 2.7.9 and Windows)

PR #89, opened 2016-03-11
Reset added by the end of flashing

igrowing:


pfalcon:
I don't know if anybody else just shy to give -1 here, I'm for sure do. But now people asking why esp-open-sdk uses a fork of this repo, and not this repo directly, and it's because of the fact that such patches are not rejected right away (and thus, have non-zero probability to be actually merged). And then I need a place to have them reverted.

So, let's see - it doesn't make sense to "reset by the end of flashing". Perhaps, in niche cases, it may make sense, and may warrant an *optional* flag to do that. But generally, there's nothing wrong to flash a dozen of binary files at different locations one by one. There's no need to reset between them. There's no need to wait 3 seconds between them.

projectgus:
Hi @pfalcon,

I'm sorry to hear that you feel this way.

> now people asking why esp-open-sdk uses a fork of this repo, and not this repo directly, and it's because of the fact that such patches are not rejected right away (and thus, have non-zero probability to be actually merged).

This patch is not going to be merged, at least in its current form. At minimum it breaks our automated coding standard check, but I also agree it needs work and some discussion. We just haven't had that discussion yet.

Would you please consider tracking this upstream esptool.py but raising issues if you see any breaking changes? You still have full control over when you update the submodule commit in esp-open-sdk. I promise that anything that turns out to be a regression or breaks backwards compatibility will be addressed.

This reduces duplication of effort, and potential confusion over there being a parallel-developed fork of this project.

> It doesn't make sense to "reset by the end of flashing". Perhaps, in niche cases, it may make sense, and may warrant an optional flag to do that.

I think there are reasons to discuss the need for an option of this kind, although neither of the current patches addressing this (here and #80) are ready for merging.

The bootloader "flash finish" command in ROM runs the firmware image, so the current write_flash behaviour includes a "soft reset" at the end.

If you run the current esptool.py multiple times in sequence then there will be soft resets to the firmware each time as the flash ends and then esptool.py re-runs (for this reason the esptool 'write_flash' command includes a provision to write multiple images in one pass.)

After the bootloader soft resets, the ESP's state is different to under a hard reset - baud rate remains set as per esptool, SPI registers remain set as per esptool. Most noticeably if the firmware issues a reset itself (for example following a fatal exception, watchdog timeout, or after an OTA upload) then it resets directly to the bootloader again.

Sometimes this behaviour is desirable, but sometimes it isn't.

@igrowing, thanks for the patch and apologies for hijacking your PR. Are you able to make this hard reset feature into an option rather than the default? Is there a reason you found that you needed a 3 second delay after the final reset? If you update the PR can you please [squash your commits](http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit/) into a single commit, as per the link? Thanks!

pfalcon:
> Would you please consider tracking this upstream esptool.py but raising issues if you see any breaking changes?

Sure, that's what I do of course.

> I promise that anything that turns out to be a regression or breaks backwards compatibility will be addressed.

Thanks, good to know. Because lately there were few things merged which I personally wouldn't merge (note: I don't say they shouldn't have been merged, just I wouldn't), so I wondered if it'll end up with any  patch submitted being merged. It would be nice to know that esptool.py will keep adhering to unix tradition of providing orthogonal set of tools which can be combined per user needs instead of simplifying particular workflow while making other harder.

> If you run the current esptool.py multiple times in sequence then there will be soft resets to the firmware each time as the flash ends and then esptool.py re-runs (for this reason the esptool 'write_flash' command includes a provision to write multiple images in one pass.)

Well, maybe you're right and what I'm actually making noise of is sleep(3) and perceived uncontrolled merging. I do remember however that I had a workflow when I had a module hardwired to boot in bootloader mode, then was able to flash thing multiple times, and when I needed to run user firmware, I used esptool.py run. It might have been with esptool.py patched to not bootloader-reset after write_flash. I generally have had a lot of local patches which I'm too shy to submit, because I'm not sure if they're my whim or generally useful, or if their changing existing behavior is acceptable. I hope a similar conservative attitude will be used by maintainers too.

pfalcon:
@projectgus : Btw, when using write_flash --verify with latest master, device remains in bootloader mode, and I need to execute esptool.py run manually.

projectgus:
Thanks for the bug report, I've opened an issue to track it. Thank you also for resetting your fork to track the upstream branch.

If you have private patches that you find useful then I would be very interested to see them, please consider submitting them as PRs even if you don't think they're suitable for merging yet. Knowing how various people use esptool is an essential step to making sure it remains useful. Also, it may happen that other people find those patches useful too - and through that broader discussion we may be able to find ways to adapt esptool in the long run.

Regarding the overall philosophy of esptool, without more specifics I don't really understand what you're driving at. If you have particular concerns then please open a fresh issue or email me offline.

pfalcon:
Ack, thanks.

projectgus:
esptool.py v2.0 adds hard reset after flash by default, with option to fall back to "soft reset" (ie run program) or disable resetting entirely.

Details:
https://github.com/espressif/esptool#reset-modes

PR #90, opened 2016-03-14
new optional --flash_block_size SIZE parameter added

jandelgado:
added a new command line parameter: "--flash_block_size SIZE" which allows

to optionally override the ESPROM.FLASH_BLOCK constant. The default behaviour
is unchanged (default value of 0x400).

The change is necessary since the default setting does not work with all
usb-to-serial adapters and can lead to error during the write_flash operation.

For my FTDI232RL board, a value of 64 allows an error free write_flash operation.


projectgus:
Are you powering your ESP8266 via the FTDI chip's internal 3.3V regulator? Other people have found this leads to brownouts (FT232 can't supply enough current at 3.3V to power the ESP8266).

Reducing the flash block size can work around this but it's not a good long term solution.

PR #92, opened 2016-03-16
Honour commandline -fs/-fm/-ff when verifying.

jmattsson:
When using `write_flash --verify`, the verify typically fails due to -fs/-fm/-ff switches used not matching the underlying file. This patch makes the verification step compare against what was actually intended to be written instead of the raw file. Plain `verify_flash` still behaves as before.

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -576,11 +576,12 @@ def write_flash(esp, args):
             print '\rWriting at 0x%08x... (%d %%)' % (address + seq * esp.ESP_FLASH_BLOCK, 100 * (seq + 1) / blocks),
             sys.stdout.flush()
             block = image[0:esp.ESP_FLASH_BLOCK]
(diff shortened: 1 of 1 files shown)


projectgus:
Thanks for this! It'd been on my mind as well since I added the --verify option, much appreciate you fixing it.

I merged it as ee00b3f3, plus a modification 219fd023 to not store state in args.

jmattsson:
Ah, even nicer, thanks! I couldn't get the parser stuff to be happy when I simply passed the block0 in as another parameter to verify_flash (can you tell I'm not a Python person? :laughing: )

projectgus:
No worries. The "magically call the module function with the same name as the command" functionality is a big weird as well, makes that part non-intuitive (and I had to change it before I could do it this way, anyhow).

PR #93, opened 2016-03-17
Use bulk reads to massively reduce overheads.

jmattsson:
This patch makes it possible to verify large firmwares from a Raspberry Pi 2 too. Previously the overhead of reading a single byte at a time (going all the way into the kernel and back) made it impossible to verify at 230400 baud rate, and at 115200 it put a core at ~90%. With this patch, a verify at 230400 is both possible and only uses ~30% cpu.

Patch by @umisef (with white-space cleanup by yours truly).

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -77,17 +77,25 @@ class ESPROM:
     def read(self, length=1):
         b = ''
         while len(b) < length:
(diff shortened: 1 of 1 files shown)


projectgus:
Thanks!

PR #95, opened 2016-03-19
Set python2 explicit

dns2utf8:
because distros are switching the default to python3 and a less eperienced user may fail to interpret the error-message.

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 #
 # ESP8266 ROM Bootloader Utility
 # https://github.com/themadinventor/esptool


projectgus:
This comes up a lot, but unfortunately it doesn't work on all platforms either. Issue #21 goes over the whole sad story.

Are you an Arch user? It seems like Arch is the distro that went against the advice in PEP394, so for now Arch users miss out :(. (I personally find this extra annoying because I'm thinking about becoming an Arch user myself!!)

The only effective long term solution that we can find is to support 2 & 3 simultaneously in the same Python file. I pushed up a branch that does this (python2_and_3) last year but it's very out of date now, would probably need someone to recreate it with recent changes before it could be merged.

There's more discussion of all this in #21.

PR #101, opened 2016-04-19
Update README.md

marcelstoer:
8 * 115'200 = 921'760

921'600 leads to `Invalid argument` in `set_special_baudrate`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ If using Cygwin on Windows, you have to convert the Windows-style name into an U
 
 ### Baud rate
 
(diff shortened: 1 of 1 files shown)


marcelstoer:
So does 921'760 and 460'800.. no clue.

projectgus:
What OS are you using and which serial adapter?

marcelstoer:
I'm on OS X and I tested flashing a WeMos D1 mini which uses a CH340 adapter from Winchiphead. 230'400 is the highest I can go. I wouldn't expect esptool to throw "Invalid argument" if a baud value is simply too high for the serial adapter.

projectgus:
The adapter can go that high, maybe the OS X driver can't.
Can you please post the full output from esptool.py, including the exception stack trace?

marcelstoer:
At this point it makes more sense to track this as an issue.

PR #103, opened 2016-04-20
Use Cesanta flasher stub for flashing.

rojer:
Faster, with better checksums.

Described in detail here:
https://blog.cesanta.com/esp8266-superfast-flashing-flash-esp-in-5-seconds

Minor change: write_flash now does not report progress by default,
--progress option has been added to enable it (same as for read_flash).

themadinventor/esptool#96

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -17,21 +17,24 @@
 # this program; if not, write to the Free Software Foundation, Inc., 51 Franklin
 # Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
-import sys
-import struct
-import serial
-import time
 import argparse
-import os
-import subprocess
-import tempfile
+import hashlib
 import inspect
+import io
+import json
+import os
+import serial
+import struct
+import subprocess
+import sys
+import tempfile
+import time
 
 
 __version__ = "1.0.2-dev"
 
 
-class ESPROM:
+class ESPROM(object):
     # These are the currently known commands supported by the ROM
     ESP_FLASH_BEGIN = 0x02
     ESP_FLASH_DATA  = 0x03
@@ -61,11 +64,8 @@ class ESPROM:
     ESP_OTP_MAC1    = 0x3ff00054
     ESP_OTP_MAC3    = 0x3ff0005c
 
-    # Sflash stub: an assembly routine to read from spi flash and send to host
-    SFLASH_STUB     = "\x80\x3c\x00\x40\x1c\x4b\x00\x40\x21\x11\x00\x40\x00\x80" \
-                      "\xfe\x3f\xc1\xfb\xff\xd1\xf8\xff\x2d\x0d\x31\xfd\xff\x41\xf7\xff\x4a" \
-                      "\xdd\x51\xf9\xff\xc0\x05\x00\x21\xf9\xff\x31\xf3\xff\x41\xf5\xff\xc0" \
-                      "\x04\x00\x0b\xcc\x56\xec\xfd\x06\xff\xff\x00\x00"
+    # Flash sector size, minimum unit of erase.
+    ESP_FLASH_SECTOR = 0x1000
 
     def __init__(self, port=0, baud=ESP_ROM_BAUD):
         self._port = serial.Serial(port)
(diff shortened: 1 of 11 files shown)


esptool.py
```diff
@@ -58,44 +63,36 @@ class ESPROM:
...
-                        pos = 0
-                    c = d[pos]
-                    pos = pos + 1
-                    if c == '\xdc':
-                        b = b + '\xc0'
-                    elif c == '\xdd':
-                        b = b + '\xdb'
-                    else:
-                        raise FatalError('Invalid SLIP escape')
+        while True:
```
projectgus: This looks like it reverts #93, which was apparently a big performance boost on slower platforms like Raspberry Pi

projectgus: ... maybe it's possible to use io.BufferedReader to keep this code simple, while keeping overhead low on slow platforms?

esptool.py
```diff
@@ -588,6 +569,192 @@ def load_section(self, section):
         return data
 
 
+class CesantaFlasher(object):
+    # This is "wrapped" stub_flasher.c, to  be loaded using run_stub.
+    FLASHER_STUB = """\
```
projectgus: Could the stub hex data be put at the very end of the file, please? I think that's the best compromise between readability and keeping everything in one source file.

esptool.py
```diff
@@ -689,43 +867,31 @@ def write_flash(esp, args):
     flash_mode = {'qio':0, 'qout':1, 'dio':2, 'dout': 3}[args.flash_mode]
     flash_size_freq = {'4m':0x00, '2m':0x10, '8m':0x20, '16m':0x30, '32m':0x40, '16m-c1': 0x50, '32m-c1':0x60, '32m-c2':0x70}[args.flash_size]
     flash_size_freq += {'40m':0, '26m':1, '20m':2, '80m': 0xf}[args.flash_freq]
-    flash_info = struct.pack('BB', flash_mode, flash_size_freq)
+    flash_params = struct.pack('BB', flash_mode, flash_size_freq)
+
+    flash_baud = args.flash_baud and args.flash_baud or 0
```
projectgus: You can achieve this by passing "default=0" to the parser.add_argument() call

flasher_stub/stub_flasher.c
```diff
@@ -0,0 +1,352 @@
...
+
+  if (last_cmd == CMD_BOOT_FW) {
+    /*
+     * Find the return address in our own stack and change it.
+     * "flash_finish" it gets to the same point, except it doesn't need to
+     * patch up its RA: it returns from UartDwnLdProc, then from f_400011ac,
+     * then jumps to 0x4000108a, then checks strapping bits again (which will
+     * not have changed), and then proceeds to 0x400010a8.
+     */
+    volatile uint32_t *sp = &baud_rate;
```
projectgus: This is a nice trick :)

esptool.py
```diff
@@ -44,6 +47,8 @@ class ESPROM:
     ESP_RAM_BLOCK   = 0x1800
     ESP_FLASH_BLOCK = 0x400
 
+    ESP_FLASH_SECTOR = 0x1000
```
projectgus: Because of the commento in line 35, it's probably worth commenting here that some commands are stub-only rather than in the ROM

flasher_stub/README.md
```diff
@@ -0,0 +1,3 @@
+This is the source of CesantaFlasher.FLASHER_STUB.
+
+Environment to compile it is available [here](https://github.com/cesanta/fnc/tree/master/common/platforms/esp8266/stubs).
```
projectgus: Do you think it would be viable to include enough additional files to compile the stub in-tree here? (Except for the 3 small SDK header files, at least.)

projectgus:
Hi @rojer,

This looks really exciting. I've left a few minor comments inline, but generally I think this is a good development - more stable and faster are both big pluses.

Because it's a large change I'd also like to test and review it a bit longer before merging. I'm about to cut a 1.0 release and I think it's sensible to put this in after that, so there's a well-tested stable release to fall back on. But generally I'm looking forward to it. :)

Thanks very much!


Angus

esptool.py
rojer: so, it's tricky. we don't know upfront how much data stub is going to send, and pyserial (at least the posix implementation) is bent on reading at least as many bytes as instructed.
even timeout stops working once at least one byte has been read.
what they do there (in PosixPollSerial.read) doesn't make sense to me and seems like a bug, but it is what it is.

however, since with this change verification does not require reading out all the bytes and performs on-device digest, the impact of going back to reading byte by byte is reduced substantially.

esptool.py
rojer: done

esptool.py
rojer: good point, done

flasher_stub/README.md
rojer: ok, done.
it still depends o docker image, but it shouldn't be a problem to remove that by adding a SDK_PATH variable or some such. i do not actually have the toolchain installed to test w/o docker, so i'll leave it to you. should be trivial.

projectgus:
I played around with the flasher stub a bit over the past couple of days, and pushed some minor changes here:
https://github.com/themadinventor/esptool/commits/stub_loader

(The SLIP syncing change also cleaned up some of the tolerance for baud rate changes, the NodeMCU I have here now flashes at 1.5Mbaud which is nice!)

Would appreciate any comments you have on those changes, if you have the time. :)

There's one more thing I was thinking about before merging, which is the --baud option. It seem fairly redundant now compared to --flash_baud. Apart from throughput performance, the only reason I know of to change it is a bad serial connection that needs to sync at less than 115200.

I'm thinking of just using `--baud` for everything, but only syncing & performing any ROM operations at max 115200bps - so higher rates only ever get used by the flasher. It seems a bit odd, but it's fully backwards compatible and I can't think of any reason someone would want different behaviour. What do you think?

rojer:
yes, makes sense. i cherry-picked the stub makefile and progress changes into this PR and got rid of --flash_baud in favor of --baud.

i think the other two changes should better be made separately so as not to increase the size of this PR, which is big enough already.

please take another look.

projectgus:
Oops, I think we doubled up - I just pushed 8031ecee90. Do you want to try to squash that commit into this PR, too? Otherwise I'll rebase that branch after this PR goes in, I agree it makes sense to split it out.

rojer:
cherry-picked and squashed.

can you please keep the "Cesanta" name in the message about stub? at least until it's substantially rewritten. a little bit of PR for us :)

projectgus:
Thank you,  merged! 

Regarding the message, no problem. I'm very happy to acknowledge Cesanta's contribution including a link on the Readme, etc. I will probably remove the name from the status message eventually, I'm mostly concerned from a usability perspective -  it'll be a new word for most users, and it's the kind of thing they're likely to Google when looking up error messages, and that may just cause further confusion. But I'm fine to keep it for now, at least - as you suggest - until the stub starts evolving. 

Cheers

PR #108, opened 2016-05-10
Make sure python 2 is used

torwag:
The script still uses python 2 syntax, more and more distros moved from python 2 to python 3 as standard python installation. Thus, they link python to python3. This makes it necessary to  explicit  call python2 for those distros.

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 #
 # ESP8266 ROM Bootloader Utility
 # https://github.com/themadinventor/esptool


projectgus:
Hi,

Thanks for sending the PR. This actually comes up once every few months, which inspired me to add the comment in the commit above.

I'm very sympathetic to this problem (I started using Arch Linux recently and they mapped python->python3 some years ago). However as discussed in #21 there are no universal solutions apart from rewriting esptool.py to support both 2 & 3.

I put some tips for Arch users here: https://github.com/themadinventor/esptool/wiki/Arch-Linux-Tips
(also linked in the other issue.)

Please let me know if more distros other than Arch (& Arch-derived) have also changed their default Python. It'd be helpful to know if it's starting to become more common.

Angus

PR #111, opened 2016-07-14
Experimental ESP31 support

projectgus:
I've been working on adding ESP31 support (as a basis for ESP32 support, once hardware is available) support to esptool.py, based on the version that Espressif [shipped in their SDK](https://github.com/espressif/ESP32_RTOS_SDK/blob/master/tools/esptool.py) with some help from a contact at Espressif.

Compared to the version in the ESP32 SDK, this version should provide a more uniform interface and is backwards compatible with ESP8266. Chip type is auto-detected from the bootloader.

# Missing Functionality
* No erase chip support (the ROM address of this function is unknown).
* No software-initiated reset after flashing, have to hard reset.
* elf2image only supports ESP8266 (ESP32 SDK uses a "gen_app.py" script to generate binaries from ELF files)

# Untested
* Baud rates above 115200 (the adapter I have doesn't work properly above 115200. I believe the baud rate code is working, because I can set baud rates below 115200 without a problem.)

# Known Issues
* ESP31 uses "synchronous" rather than "asynchronous" flash writing method, the interrupt management functions aren't exposed in ROM so we can't use an ISR. This means the performance boost from using a flasher stub program is limited. I believe ESP32 will have these ROM functions documented so we can go back to async loading, like on ESP8266. This should simplify the stub code as well.
* MD5 is also not documented in the ESP31 ROM, so it's compiled into the ESP31 stub. Hoping this will also not be needed for ESP32.

Please report any bugs or results found during testing, as a comment on this PR.

It is likely that ESP31 support will not be long-lived in esptool.py, and once ESP32 is released we will support ESP8266 & ESP32 only. This is because ESP31 hardware is prerelease beta only, very limited numbers, and also because the ESP32 stub will need changing from ESP31, so keeping ESP31 support will be quite a lot of additional work.


projectgus:
Turns out supporting ESP31 is a significant amount of work compared to ESP32, so it will not be supported in future versions. See #121 for ESP32 support.

PR #119, opened 2016-08-12
Update esplool.py to use python2 not python3

regimantas:
After install something to my linux system default python2 changed to python3 and wen i try using esptool it give me grammatical error. Changing #!/usr/bin/env python to #!/usr/bin/env python2 solve problem :)


projectgus:
Hi,

Thanks for your concern with improving esptool.py. Unfortunately, this problem is complex. Please read the comment on the line directly under the line you changed.

Angus

PR #124, opened 2016-09-09
Warn when data does not fit in flash

jms19:
People flashing images particularly Micropython which has recently passed 512k in size using the default flash size are upset when esptool appears to work but in reality doesn't work as intended.

At a minimum warn the user about it. Perhaps it should stop there and then instead.


deshipu:
esptool.py:836:74: E227 missing whitespace around bitwise or shift operator

projectgus:
Thanks, this looks like a good addition. I had something similar on the checklist for v2.0 #121 . v2.0 will probably fail in this case, but warning like this is a good stop-gap.

PR #126, opened 2016-09-18
esptool: use serial_for_url instead of Serial

hoihu:
This change allows ports to be opened locally and remote (for example using RFC2217, but also other messaging protocols may be used).

see http://pyserial.readthedocs.io/en/latest/pyserial_api.html#serial.serial_for_url for details

It's useful if the compiling instance does not have direct serial port access to deploy the firmware (e.g. if you are within a docker container). 

serial_for_url is backward compatible to the existing Serial() call.

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -68,7 +68,7 @@ class ESPROM(object):
     ESP_FLASH_SECTOR = 0x1000
 
     def __init__(self, port=0, baud=ESP_ROM_BAUD):
-        self._port = serial.Serial(port)
+        self._port = serial.serial_for_url(port)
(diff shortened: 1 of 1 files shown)


projectgus:
Hey,

Sorry I never got back to you on this. This is really cool, I had not noticed pyserial can do this!

Happy to merge, the only thing I noticed is that serial_for_url was new pyserial 2.5. Previously I think older versions would work.

Version 2.5 is 6 years old (in Debian since Wheezy, in Ubuntu since Precise) so I'm happy to add this requireut. But could you please also add a note to the existing README section about pyserial, mentioning 2.5 or newer? Thanks.

PR #127, opened 2016-09-22
port esptool to python 2/3

jeremyherbert:
I've ported esptool to python 3.

I tried to change as little as possible, because the refactoring probably should come in a different commit. I suggest in the future that all byte strings be removed, and there is a transition to passing data around as a list of integers.

The only functional change I made (I think ;) ) was that there was a blind try/except which was hiding exceptions in the early part of the connection routine, so it now prints out the exception string.

I have tested the following things to work with both python 2.7.5 and python 3.5.1 on a nodemcu v0.9 by flashing micropython:

```
python esptool.py -p /dev/tty.wchusbserial1420 write_flash --flash_size=8M --verify 0 ~/Downloads/esp8266-20160909-v1.8.4.bin

python esptool.py -p /dev/tty.wchusbserial1420 read_mac

python esptool.py -p /dev/tty.wchusbserial1420 chip_id

python esptool.py -p /dev/tty.wchusbserial1420 flash_id

python esptool.py -p /dev/tty.wchusbserial1420 erase_flash
```

I'm not all that experienced with the other features of esptool, so I would appreciate some help in testing them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ sudo: false
 
 python:
   - "2.7"
+  - "3.4"
   - "3.5"
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ sudo: false
 
 python:
   - "2.7"
+  - "3.5"
 
 script:
   - python setup.py build
--- a/esptool.py
+++ b/esptool.py
@@ -18,6 +18,8 @@
 # this program; if not, write to the Free Software Foundation, Inc., 51 Franklin
 # Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
+from __future__ import print_function, unicode_literals, division
+
 import argparse
 import hashlib
 import inspect
@@ -78,33 +80,37 @@ class ESPROM(object):
 
     """ Read a SLIP packet from the serial port """
     def read(self):
-        return self._slip_reader.next()
+        return next(self._slip_reader)
 
     """ Write bytes to the serial port while performing SLIP escaping """
     def write(self, packet):
-        buf = '\xc0' \
-              + (packet.replace('\xdb','\xdb\xdd').replace('\xc0','\xdb\xdc')) \
-              + '\xc0'
+        buf = b'\xc0' \
+              + (packet.replace(b'\xdb',b'\xdb\xdd').replace(b'\xc0',b'\xdb\xdc')) \
+              + b'\xc0'
         self._port.write(buf)
 
     """ Calculate checksum of a blob, as it is defined by the ROM """
     @staticmethod
     def checksum(data, state=ESP_CHECKSUM_MAGIC):
         for b in data:
-            state ^= ord(b)
+            if type(b) is int:  # python 2/3 compat
+                state ^= b
+            else:
+                state ^= ord(b)
+
         return state
 
     """ Send a request and read the response """
     def command(self, op=None, data=None, chk=0):
         if op is not None:
-            pkt = struct.pack('<BBHI', 0x00, op, len(data), chk) + data
+            pkt = struct.pack(b'<BBHI', 0x00, op, len(data), chk) + data
             self.write(pkt)
 
         # tries to get a response until that response has the
         # same operation as the request or a retries limit has
         # exceeded. This is needed for some esp8266s that
         # reply with more sync responses than expected.
-        for retry in xrange(100):
+        for retry in range(100):
             p = self.read()
             if len(p) < 8:
                 continue
@@ -119,15 +125,15 @@ class ESPROM(object):
 
     """ Perform a connection test """
     def sync(self):
-        self.command(ESPROM.ESP_SYNC, '\x07\x07\x12\x20' + 32 * '\x55')
-        for i in xrange(7):
+        self.command(ESPROM.ESP_SYNC, b'\x07\x07\x12\x20' + 32 * b'\x55')
+        for i in range(7):
             self.command()
 
     """ Try connecting repeatedly until successful, or giving up """
     def connect(self):
-        print 'Connecting...'
+        print('Connecting...')
 
-        for _ in xrange(4):
+        for _ in range(4):
             # issue reset-to-bootloader:
             # RTS = either CH_PD or nRESET (both active low = chip in reset)
             # DTR = GPIO0 (active low = boot to flasher)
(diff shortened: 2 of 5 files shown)


jeremyherbert:
Also: I've only tested this on my mac. Python 2 was the system install, python 3 was an anaconda install. Any other operating system testing would be appreciated.

jrobeson:
it's just gonna get rotten again if the maintainers don't commit to always using `b'` on byte strings

esptool.py
```diff
@@ -18,6 +18,8 @@
 # this program; if not, write to the Free Software Foundation, Inc., 51 Franklin
 # Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
+from __future__ import print_function, unicode_literals, division
+
```
jrobeson: I'm not so sure about unicode literals. I don't know the codebase well enough, but in a previous 2/3 compat project, we ran into issues with lots of os level APis not accepting unicode, so I'd be careful with it.  Some more discussion is here: https://github.com/PythonCharmers/python-future/issues/22

jrobeson: If future is going to be required, then it should be part of the flake8 configuration too, so it'll reject any new files without these future headers.

.travis.yml
```diff
@@ -3,6 +3,7 @@ sudo: false
 
 python:
   - "2.7"
+  - "3.5"
```
jrobeson: if we'd like be sure that it works out of the box on current debian stable and up, then 3.4 should be added too.

jeremyherbert: done. thanks.

esptool.py
jeremyherbert: added to the flake8 config. As for whether to use unicode_literals or not, I will defer to the big boss @themadinventor as to what he thinks

projectgus:
Hi @jeremyherbert & @jrobeson,

This looks like good work, thanks for sending it! There are a couple of thoughts I have on this:

* Re: unicode literals, what's the upside to using them? Just that it forces us to never assume any non-byte-string is 8 bits wide?

* I have a Work-In-Progress v2.0 PR open in #121. This adds some very major refactors, including ESP32 support. It also adds automated tests, similar to the set of tests you ran above. So although we can't run those on-device tests via Travis, we can run them manually (and eventually I'd like to integrate them with github via an external CI tool).

Which means there are two possible avenues here:

* Do a V1.2 release which supports Python 2 & 3, followed closely by a v2.0 release which will also support both.

* Roll up the Python 2 & 3 support as yet another V2.0 feature.

If we put it into V2.0 and run automated tests on both Py2 & 3, I think we can probably live without unicode literals - the tests should "keep us honest". :)

I'm leaning towards the second option but I could be convinced to do a V1.2 release as well.

jrobeson:
I didn't do anything, but I am against `unicode_literals`. I'd rather see explicit `u'` where unicode is allowed.

jeremyherbert:
@projectgus here is my opinion on these things, but ultimately it's not my project :)

I think there should be a v1.2 with only Python 2/3 support. I think that releases should be as small as possible, and adding esp32 at the same time makes it hard to work out where something went wrong in the past. 

On the Unicode literals, I think they should stay. I think that people should get used to all strings being Unicode in Python, and everything should be designed with Python3 as the primary target, with python2 added in a secondary fashion. Most of the PyCon talks by core developers in the last few years seem to have one slide where they just beg people to stop using Python2. Some Linux distributions don't even ship python2 anymore (and others are trying to join them). Adding the u'' everywhere is redundant and unnecessary in Python3, so I wouldn't want to add it in. 

The only people who should be actively using python2 still are perhaps those with old, large codebases and complex C extensions. This project targets modern chips (esp32 is barely out yet!), has a very small code base and no compiled extensions, so I don't think it makes sense to go out of the way to keep it. 

Frankly, if it was up to me I would just drop python2 altogether ;)

jrobeson:
I don't like using the explicit `u'` either, but it's the only reasonable way forward (imo) if 2/3 support is to be kept.  

@projectgus doesn't just have to worry about about folks who use esptool directly, but also downstreams that might have built on top of it.  It's not a decision to make lightly.

jeremyherbert:
Again, this is just my opinion, but I would say release 1.2 with python 2/3 support, and say that it is the last version with python 2 support. Other large projects like scons are taking this approach. People can still use the version if they want python 2 for whatever reason. 

Then, add the esp32 support to the python3-only version (v2 onwards). You won't break any existing tools, because there aren't any yet :)

projectgus:
Hi everyone,

Discussed this with @jrobeson last week on IRC. Forgot to update here.

* For V1.2 release, I think it's easiest to not use `u""` or `unicode_literals`, and have strings be their "native" type on py2 or py3.
* For the future V2.0 releases, "subprocess" is removed so it should be much easier to use `unicode_literals` from there on forward.

Regarding moving Python 3 only, V2.0 was probably the perfect release to do this in, but we're already shipping esp-idf to people with Python 2 & the V2.0 development version. So Python 2 support is probably going to stick around for a while longer. :( 

For V2.0 I will update all the relevant documentation to encourage everyone to use the Py 3 version, and see how things go from there I guess...

jrobeson:
so would this PR be merged if unicode_literals was removed?

projectgus:
Yes, I think that & resolving the merge conflict is all that's needed.

I'll probably do some of my own tests locally before merging (I can backport the test script I use in v2.0), but I expect it'll be good to go.

jeremyherbert:
Ok, I will try to get on to changing this on the weekend (unless someone else beats me to it)

projectgus:
Thanks @jeremyherbert , much appreciated. :)

projectgus:
@jeremyherbert , I snuck in a couple of other v1.2 commits before this one. If you get a chance to unpick the merge conflict that would be great, thank you. Let me know if you don't.

Once this is merged I'll run some more tests and make the v1.2 release on pypi.

jeremyherbert:
I don't think it will be an issue

projectgus:
Bumped this to a new v1.3 milestone. If anyone here has the time to resolve the merge conflict, that would be great. Otherwise I'll try to make time soon.

jrobeson:
@projectgus : so we talked about whether the interface would be compatible between esptool 1.x and 2.x.   Did you decide against doing so?

projectgus:
@jrobeson I have not yet decided. The v2.0-dev interface is still compatible at the moment, but there are some things I'd like to change in it. I'll keep it fully backwards compatible if it's not too awkward, though.

I am still committed to doing a v1.3 release with Python 3 support but without the ESP32 changes/features.

jeremyherbert:
Hi again folks, 

Sorry for the delay on my end. I've been a bit stuck with real life at the moment (in the middle of writing up my phd thesis). I will get to it as soon as I can, but don't wait up for me if you think you can do it yourself. Sorry.

projectgus:
Closing this in favour of #150. Comments/review/additions on that PR welcome!

PR #128, opened 2016-09-26
Use Cesanta stub for erase_flash

jimparis:
This has the benefit of letting us wait until it's finished, unlike `esp.flash_erase()`.  As a result we can now report how long it takes.  For reference, it takes 22.4 seconds on my ESP-WROOM-02.


projectgus:
Thanks @jimparis, this looks great. We have this feature in the v2.0 development branch (#121), but it's good to have it in master as well.

PR #131, opened 2016-10-08
write_flash: Default to auto-detect Flash size based on ID.

pfalcon:
If --flash_size=detect (new default) is given, a FlashID will be read, and
Flash size to put into a firmware header will be calculated based on it
(one subfield of FlashID is power-of-2 of the chip capacity, and most modern
chips follow this convention).

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ The --port argument specifies the serial port. This may take the form of somethi
 
 The next arguments to write_flash are one or more pairs of offset (address) and file name. When generating "version 1" images, the file names created by elf2image include the flash offsets as part of the file name. For "version 2" images, the bootloader and linker script you are using determines the flash offset.
 
-You may need to specify arguments for [flash mode and flash size](#flash-modes) as well. For example:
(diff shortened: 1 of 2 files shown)


pfalcon:
This implements https://github.com/themadinventor/esptool/issues/105.

projectgus:
Hi @pfalcon,

This looks great, thanks. A big usability improvement.

The only thing I'm unsure about is throwing a fatal error if the flash size can't be detected. I'd prefer printing a warning and falling back to the old default. This way if anyone is using the default flash_size successfully now with unusual flash, an update won't change their behaviour.

Angus

projectgus:
One other thing, the README needs updating to mention the new argument as well. We can probably remove --flash_size from the "default" write_flash example as well, which is great.

pfalcon:
Updated based on the comments above.

projectgus:
Awesome, thanks!

pfalcon:
Thanks for merging. Any chance this can be released on PyPI sooner rather than later?

projectgus:
I'm planning a v1.2 release, there's one remaining PR assigned to this release (link: https://github.com/themadinventor/esptool/milestone/1 ). So sooner seems likely.

pfalcon:
Thanks.

PR #152, opened 2016-12-19
Added support for partial flash erase.

tai:
I have added a code to erase just a part of the flash memory.
I needed this to support multi-boot configuration that allows you to switch between 4 1MB ROMs on 32Mbit Flash.

With this patch, CLI syntax is enhanced so you can type

`esptool erase_flash <start-address> <length>`

Or, by example, you can type

`esptool erase_flash 0x2000 0xfe000`

and erase 1st ROM only, preserving the boot loader and other (2nd to 4th) ROMs intact (NOTE: address and length depends on actual ROM layout). Without start address and length supplied, it will simply erase whole flash region, which is a prior behaviour.

Please consider applying this patch, as this improves productivity when you are working with multi-ROM configuration.

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -551,6 +551,7 @@ class ELFFile(object):
 class CesantaFlasher(object):
 
     # From stub_flasher.h
+    CMD_FLASH_ERASE = 0
     CMD_FLASH_WRITE = 1
     CMD_FLASH_READ = 2
     CMD_FLASH_DIGEST = 3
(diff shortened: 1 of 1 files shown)


projectgus:
Thanks for sending this, @tai. I agree it's a useful feature. It's actually present in the v2.0 development version (the branch corresponding to #121) as the commmand "erase_region".

At the moment I'm trying to finalise a v1.3 stable release and then merge v2.0 to master to get past the current split development state, which is a bad situation (not least because it leads to doublings-up of effort like this!)

At the moment my preference is not to merge this, in order to make it easier to get v1.3 released and then merge v2.0 to master - at which point the functionality will be available via erase_region. This should happen in the next week or two. I'm really sorry for the hassle.

projectgus:
The master branch is now esptool v2.0, which has the new "erase_region" command.

If you think any functionality is missing from this command, please let me know (or feel free to send another PR).

PR #162, opened 2017-01-03
Verify: Skip verification of first four bytes at address 0

thespooler:
Since the bytes 2-3 contain flash_params, they can't be verified afterwards.
Changed default behavior to verify and added --no-verify switch.


projectgus:
Implemented in bbdeb6683428368bd8d0f83f2d3f593f1cf1f221 using a different approach.

Issue #146, opened 2016-11-24
Address - Binary pair issue


forrestfiller:
I'm less of an advanced user, apologies if this is too basic of a question:

I've been struggling for the past day with flashing my esp, as I'm unsure what they mean by address? I have used 0x00000 as well as 0x0 before my .bin file. Have a look at the enclosed screenshot. Thanks for your time.

- OS: running VirtualBox on Sierra, using an Ubuntu 16 64-bit image
- ESP2866 flash id: e0, 4016 ( which is a A25LQ032) "Huzzah" breakout from Adafruit Industries (chip id: 0x000e363a)
- powered through my laptop's usb port, with a TTL/USB cable

![virtualbox_ubuntu16_24_11_2016_10_00_46](https://cloud.githubusercontent.com/assets/18606441/20602907/39038a1c-b22d-11e6-9a5a-ad827ae7056d.png)

projectgus:
This is a bad error message, so it's absolutely not your fault that it doesn't make sense. The esptool options have to passed in a certain order - a small number apply to "esptool" and some apply to the "write_flash" subcommand.

This is annoying, and I've been trying to think of a good way to fix it but at the moment esptool relies fairly heavily on python's built-in argument parsing support which is what does this.

You should find this argument order works (`-b` before `write_flash`, `-p` goes before `write_flash` also.)

```
esptool.py -b 74880 write_flash -fm dio -fs 4m -ff 80m 0x0 nodemcu-dev-13-modules-2016-11-24-01-56-58-integer.bin
```

(BTW, `-fs 4m` is only 4 megabits which is half a megabyte. If you have a huzzah you probably want `-fs 32m` for 32 megabits ie 4 megabytes. Or with esptool v1.2 or newer you can leave it blank and esptool will detect the actual flash size.)

Please let me know if this works for you. In any case, I'd like to keep this issue open as a reminder that the underly argument order problem needs to be taken care of (ideally, you should be able to put -b anywhere you like. Or, at minimum, the error message you should tell you what you did wrong.)

forrestfiller:
@projectgus Thanks very much, I was able to flash the unit correctly thanks to your thoughtful and helpful reply. 

[...]

PR #173, opened 2017-01-28
Fix typos in README.

solemnwarning:
This tripped me up for longer than I'd like to admit.


projectgus:
Ouch, I'm sorry @solemnwarning ! Thanks for the fix.

There's another option issue, #146, for removing this arbitrary ordering requirement. Unfortunately Python's built-in `argparse` library doesn't make this easy, at least not that I can find. But will figure out a way.

forrestfiller:
Thank you for your efforts!

PR #176, opened 2017-02-08
Fix read_mac function for the ESP32

danicampora:
The current read_mac function in esptool is not returning the same MAC address as: ```esp_wifi_get_mac(WIFI_IF_STA, ...)```. It's only returning the last 4 bytes of the MAC address combined with the CRC at the end.


projectgus:
Oops, above commit references wrong PR number.

This looks good, thanks @danicampora . Will merge.

Issue #181, opened 2017-02-28
Install esptool.exe on Windows


hickford:
Hi. Installing esptool on Windows `pip install esptool` installs a Python script `C:\Python36\Scripts\esptool.py`. It would be more convenient if it installed an .exe file `esptool.exe`

Other applications do this such as `pip install microrepl` https://github.com/ntoll/microrepl

CharlesScoville:
Eh... exactly why would an .exe be more convenient than a .py _if you already have Python installed_? 

It's not like it's a graphics or cryptography primitive that needs to be optimized for speed, nor is there a compatibility issue since this runs on (the platform independent) Python interpreter. 

Esptool is a command-line utility. Performance is bound by the embedded devices themselves and cross compatibility ensured by Python.  It doesn't make sense to burden the developers with an extra step for no gain.

. . .

On the other hand, if you're asking for the _total package_ to be a simple/single .exe, then you should know it's already been done... and then some. Check out nodemcu-pyflasher. You can get said .exe here.

https://github.com/marcelstoer/nodemcu-pyflasher/releases

hickford:
Hi. Pardon me, I haven't explained well. The boon is the `.exe` wrapper will always execute with the right version of Python (consider if you had two copies of Python and esptool.py in your path). More at https://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation

The change is as simple as one line to setup.py, setuptools creates the `.exe` wrapper at install time https://packaging.python.org/distributing/#scripts . Let me prepare a pull request.

There might be other benefits on other platforms, I don't know

projectgus:
> The boon is the .exe wrapper will always execute with the right version of Python (consider if you had two copies of Python and esptool.py in your path).

FWIW, I think installing via the existing `scripts` directive will do this as well. If you install via `setup.py install` or `pip`, then the actual `esptool.py` file is a thin wrapper that calls the correct verison of Python and sources the script. At least that's the case on POSIX, I'd expect the same on Windows but maybe not.

(Of course this isn't the case if you run esptool.py as-is in the git repository or another standalone place. Which is one of the reason why the docs encourage a pip install.)

In any case, thanks for sending the PR. I have some more specific thoughts there, will comment there instead.

projectgus:
> I'd expect the same on Windows but maybe not.

Ah, I remember now. Windows won't execute a `.py` file properly at all, will it? You have to invoke `python` each time. Hmm.

(Left more specific comments on the PR.)

[...]

PR #184, opened 2017-03-05
In setup.py specify console_scripts rather than scripts (better cross…

hickford:
…-platform compability). See https://packaging.python.org/distributing/#scripts. Fixes https://github.com/espressif/esptool/issues/181

Tested on Windows

The principal benefit is that on Windows the application will be installed as `esptool.exe` rather than `esptool.py`

### DIFF
--- a/espefuse.py
+++ b/espefuse.py
@@ -498,9 +498,14 @@ def _get_efuse(efuses, efuse_name):
     return [e for e in efuses if efuse_name == e.register_name][0]
 
 
-if __name__ == '__main__':
+def _main():
     try:
         main()
     except esptool.FatalError as e:
         print('\nA fatal error occurred: %s' % e)
         sys.exit(2)
+
+
+if __name__ == '__main__':
+    _main()
+
--- a/espsecure.py
+++ b/espsecure.py
@@ -371,9 +371,13 @@ def main():
     operation_func(args)
 
 
-if __name__ == '__main__':
+def _main():
     try:
         main()
     except esptool.FatalError as e:
         print('\nA fatal error occurred: %s' % e)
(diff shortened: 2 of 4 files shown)


projectgus:
Thanks for sending this. I wasn't actually aware this feature existed!

This seems like the right way to do things, but I have one major concern around backwards compatibility. Essentially every guide, howto, makefile, script, etc. that runs esptool.py runs it as `esptool.py` not `esptool`.  This PR will change that naming. Also, there's at least [one other active project](https://github.com/igrr/esptool-ck), and I think at least one older project, that called itself esptool. Up until now it was possible to have both installed simultaneously without a conflict.

V2.0 changes some other behaviour of esptool.py v1.x, but it hasn't made any major breaking changes of this kind. This would be the first.

A partial workaround may be to install as scripts and as console_scripts, so both forms (`esptool` and `esptool.py`) will work. Or to at least do that on Windows, where the downsides of naming the script `esptool.py` are more obvious.

This might be too out there, but installing `esptool.py.exe` and friends on Windows might also keep things fully backwards compatible... But seems a bit icky.

projectgus:
Actually, the more I think about it specifying the console_scripts names as `esptool.py`, etc. might be the best compromise. Keen to know what you think.

hickford:
I understand your concern about backwards compatibility.

Just tried out your suggestion `esptool.py=esptool:_main` that solves my problems on Windows. Let me update the pull request.

hickford:
Updated pull request. How about that?

projectgus:
Sorry for the long delay, got busy with other things and couldn't get back to esptool. This looks great, cheers.

Issue #211, opened 2017-06-22
espsecure.py and espefuse.py traceback when no argument given


hroncok:
esptool version: 2.0
Python 3.6 on Fedora 26 (Linux)

```console
$ espsecure.py
espsecure.py v2.0
Traceback (most recent call last):
  File "/usr/bin/espsecure.py", line 385, in <module>
    _main()
  File "/usr/bin/espsecure.py", line 378, in _main
    main()
  File "/usr/bin/espsecure.py", line 372, in main
    operation_func = globals()[args.operation]
KeyError: None


$ espsecure.py foo
usage: espsecure [-h]
                 {digest_secure_bootloader,generate_signing_key,sign_data,verify_signature,extract_public_key,digest_private_key,generate_flash_encryption_key,decrypt_flash_data,encrypt_flash_data}
                 ...
espsecure: error: argument operation: invalid choice: 'foo' (choose from 'digest_secure_bootloader', 'generate_signing_key', 'sign_data', 'verify_signature', 'extract_public_key', 'digest_private_key', 'generate_flash_encryption_key', 'decrypt_flash_data', 'encrypt_flash_data')


$ espefuse.py 
espefuse.py v2.0
Traceback (most recent call last):
  File "/usr/bin/espefuse.py", line 536, in <module>
    _main()
  File "/usr/bin/espefuse.py", line 529, in _main
    main()
  File "/usr/bin/espefuse.py", line 513, in main
    operation_func = globals()[args.operation]
KeyError: None


$ espefuse.py foo
usage: espefuse [-h] [--port PORT] [--before {default_reset,no_reset,esp32r1}]
                [--do-not-confirm]
                {dump,summary,burn_efuse,read_protect_efuse,write_protect_efuse,burn_key,set_flash_voltage}
                ...
espefuse: error: argument operation: invalid choice: 'foo' (choose from 'dump', 'summary', 'burn_efuse', 'read_protect_efuse', 'write_protect_efuse', 'burn_key', 'set_flash_voltage')
```

hroncok:
Will send fix.

PR #212, opened 2017-06-22
Display usage information when no arguments are provided

hroncok:
Previously a traceback was shown instead.

Fixes https://github.com/espressif/esptool/issues/211


projectgus:
Looks great, thank you @hroncok !

PR #229, opened 2017-09-17
Fix assignment to args.chip.

Cabalist:
Fix a simple typo.  I am unsure if it actually caused any issues...


projectgus:
Good catch. I'm surprise PyFlakes didn't see this.

I think you're right it doesn't change the function behaviour, but still worth fixing.

PR #230, opened 2017-09-17
Add missing testing dependency

Cabalist:
I was missing this package in order to run all the tests.  `pyelftools` is needed by `test/test_imagegen.py`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -103,12 +103,12 @@ setup(
     tests_require=[
         'flake8>=3.2.0',
         'flake8-future-import',
+        'pyelftools',
     ],
     install_requires=[
         'pyserial>=2.5',
         'pyaes',
         'ecdsa',
-        'pyelftools',
     ],
     scripts=scripts,
     entry_points=entry_points,
--- a/setup.py
+++ b/setup.py
@@ -108,6 +108,7 @@ setup(
         'pyserial>=2.5',
         'pyaes',
         'ecdsa',
+        'pyelftools',
     ],
     scripts=scripts,
     entry_points=entry_points,


projectgus:
Thanks for bringing this up. I think I remember struggling with this whenI  was writing setup.py

I don't want to have everyone using esptool install a package that's only needed for tests. It may be enough to add it to tests_requires instead of install_requires, but IIRC there is some other catch here...

Cabalist:
Oops.  I had meant to add that to `tests_require` instead of `install_requires`.    I'm curious on what the catch is. I've recently been diving more into packaging and it can be hard to isolate edge cases unless you run into them face first.  😃 

Unrelated but is there a reason that `pyaes` and `ecdsa` are bundled AND listed as dependencies?  Seems like it would be possible to have multiple versions installed in an environment.

projectgus:
> I'm curious on what the catch is.

I am curious as well because I can't remember! I'll try and make time to play around with this change before merging and see if I can reproduce anything.

> is there a reason that pyaes and ecdsa are bundled AND listed as dependencies?

Yeah, it's a bit sub-optimal. The reason is that a lot of people have traditionally run esptool.py by git cloning and running in-place (this is also how we provide it in ESP-IDF). For a long time this was the only way to use esptool.py (now releases are put on pypi). So for now I haven't wanted to break away from the "clone the repo and it just works" method of use.

However, I also think vendoring libraries should be done as a last resort.

So, the behaviour is:
* If you "pip install" or "setup.py install" esptool.py then esptool.py will be on your PATH and will use the versions of pyaes & ecdsa installed via Python's module management system (the vendored libraries are not including in the esptool Python package).
* If you "git clone" and run it in place then you'll get the vendored libraries.

Because the pyaes & ecdsa versions are not specified in setup.py, there is the possibility that some future version of one of these libraries will break it but only for people who use the packaged version. This is sub-optimal, but I think it's also unlikely as we don't use a lot of the functionality of either library (and both seem to be mature).

Cabalist:
Deps:
Very curious here.  I wonder if it is Setuptools version dependent.  Or possibly interface dependent.  It's only setuptools' built-in test command that does anything explicitly with the test_requires setup argument. (i.e. python setup.py test)

Bundling:
Makes sense!
I'd be curious about the strategy to depreciate that.  Maybe make a branch for PyPI without those two packages and then set it to master after some time and warnings.  Afterwards maintain a branch specifically for cloning to see how many people actually do that versus `pip install`?

projectgus:
>  I'd be curious about the strategy to depreciate that. Maybe make a branch for PyPI without those two packages and then set it to master after some time and warnings. Afterwards maintain a branch specifically for cloning to see how many people actually do that versus pip install?

This could work. As a first step we would need to migrate ESP-IDF itself to use 'pip' for Python packages, so we can start to "take our own medicine" in this sense. :)

Ivoz:
How does ESP-IDF obtain pySerial?

projectgus:
Hi @Cabalist,

Digging up a very old PR. This has been fixed in master now by using the `[dev]` extras option of setup.py, `pip install .[dev]` or similar will get all the "developer recommended" packages including the ones for running tests. Hopefully this will meet this requirement from now on.

Regarding the discussing about vendored vs non-vendored packages, in esptool v3.0 we're properly embracing pip install as the way to install esptool. The last vendored package is ecdsa and will be removed soon, will close this PR when it's done.

PR #231, opened 2017-09-17
inspect.getargspec() is depreciated for Py3. 

Cabalist:
Both these functions return a named tuple as well.


projectgus:
Neat, thanks!

PR #232, opened 2017-09-17
Typo in variable name that can cause NotDefined exception

Cabalist:


projectgus:
Good catch, thanks! I think I must be relying on PyFlakes too much for catching these.

PR #239, opened 2017-10-24
correct espsecure.py command name

milan-kupcevic:
correct command name for additional help


projectgus:
Thanks @milan-kupcevic ! Will also fix the (unrelated) build failure. :)

Issue #236, opened 2017-10-04
A fatal error occurred:failed to connect to espressif Timed out waiting for packet header


Akriti-S:
Hi,

I am getting this error when I am trying to upload the sketch after uploading the SPIFFS data folder in my esp8266. 

If I upload the code without uploading the sketch's data folder then I do not get this error however if I first upload Sketch Data folder and then try uploading the sketch. I get that error. Is there any way to get rid of it.

![2017-10-04 18_35_23-mail](https://user-images.githubusercontent.com/12951630/31177228-f70b538a-a932-11e7-9688-27afb5e8adfb.jpg)

Thanks,
Akriti

projectgus:
Hi Akriti,

What ESP8266 board or hardware are you using?

Also, what Arduino IDE are you using to upload? I believe the ESP8266 Arduino IDE uses esptool-ck not esptool.py. esptool.py was (I thought) only used in the ESP32 Arduino IDE.

Angus

Akriti-S:
Hi, I am using Arduino 1.8.4. I am using nodemcu 1.0 ESP-12E. I upload the code by using esptool.py generally

projectgus:
Hi @Akriti-S ,

It looks like your dev board is failing to auto-reset into bootloader mode. You can try holding down the FLASH button while esptool.py tries to connect. If this doesn't work you can also try trapping RST while holding FLASH, while esptool.py tries to connect.

> If I upload the code without uploading the sketch's data folder then I do not get this error however if I first upload Sketch Data folder and then try uploading the sketch. I get that error. 

That's pretty odd, although it may also be related to auto-reset working once and then failing. Do you know what files are being flashed at which offsets? Is the flash size set correctly for your board?

projectgus:
Closing due to a lack of updates, but feel free to reopen if you'd like to continue the discussion about this configuration.

Issue #213, opened 2017-06-22
Setting RTC.ALARM0 on esp8266 prevents flashing


matthewbryancurtis:
Hey there, I'm having a little trouble getting esptool to erase the flash on my NodeMCU Amica development board: 

```
esptool.py --port /dev/tty.SLAB_USBtoUART erase_flash
esptool.py v2.0
Connecting........_
Detecting chip type... ESP8266
Chip is ESP8266
Uploading stub...
Running stub...
Stub running...
Erasing flash (this may take a while)...

A fatal error occurred: Timed out waiting for packet content
```

```
esptool.py --port /dev/tty.SLAB_USBtoUART --baud 115200 write_flash --flash_size=detect 0 esp8266-20170612-v1.9.1.bin 
esptool.py v2.0
Connecting........_
Detecting chip type... ESP8266
Chip is ESP8266
Uploading stub...
Running stub...
Stub running...
Configuring flash size...
Warning: Could not auto-detect Flash size (FlashID=0x0, SizeID=0x0), defaulting to 4MB
Flash params set to 0x0040
Compressed 598432 bytes to 390604...

A fatal error occurred: Timed out waiting for packet content
```

```
esptool.py --port /dev/tty.SLAB_USBtoUART --baud 9600 write_flash --flash_size=detect 0 esp8266-20170612-v1.9.1.bin 
esptool.py v2.0
Connecting....
Detecting chip type... ESP8266
Chip is ESP8266
Uploading stub...

A fatal error occurred: Timed out waiting for packet header
```

Any tips that might help me get through this issue? It also might be related to the code I put on the NodeMCU (I think I may have bricked it):

```
# boot.py
import machine

if machine.reset_cause() == machine.DEEPSLEEP_RESET:
    print('woke from a deep sleep')
else:
    print('power on or hard reset')

# configure RTC.ALARM0 to be able to wake the device
rtc = machine.RTC()
rtc.irq(trigger=rtc.ALARM0, wake=machine.DEEPSLEEP)

# set RTC.ALARM0 to fire after 10 seconds (waking the device)
rtc.alarm(rtc.ALARM0, 10000)

# put the device to sleep
machine.deepsleep()

```

Thanks for looking!

projectgus:
It does seem likely that the RTC alarm is starting and is not being disabled when esptool.py resets the board, so when it fires 10 seconds into flashing the whole thing blows up.

This is a bug we should fix, but I think you can un-brick your module right now as well:
[...]

PR #247, opened 2017-11-14
Fix timed out waiting for packet header/content error

rmokerone:
For some serial chip drivers, can not be read immediately after write.I used several chips other than cp2012 for testing, all of which were stable downloading and erasing.
This may be used to fix the following issues
#213 
#236 
etc.

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -160,6 +160,9 @@ class ESPLoader(object):
     # The number of bytes in the UART response that signify command status
     STATUS_BYTES_LENGTH = 2
 
+    # Time delay between command write and read.
+    COMMAND_WRITE_READ_DELAY = 0.3
+
     def __init__(self, port=DEFAULT_PORT, baud=ESP_ROM_BAUD):
         """Base constructor for ESPLoader bootloader interaction
 
@@ -246,6 +249,9 @@ class ESPLoader(object):
 
         if not wait_response:
             return
+
+        # the chip needs reaction time
+        time.sleep(COMMAND_WRITE_READ_DELAY)
 
         # tries to get a response until that response has the
         # same operation as the request or a retries limit has


projectgus:
Hi @rmokerone ,

This is an interesting fix, thanks for sending it. I have a couple of questions so I can understand this fix properly:

* What OS are you using?
* Is the CP2102 the only USB/serial adapter adapter where you've seen this fix a problem? Can you please explain what hardware (dev board, etc) you were using where this fixed a problem?
* How did you come up with the 0.3s (300ms) value? 300ms is quite a long time, especially relevant to USB full speed packet times. Using this delay will slow down some esptool.py operations substantially. Do smaller delays work?

rmokerone:
- macOS 10.12.6
- My English is not good, in addition to cp2102 did not encounter this problem, other chips, such as CH340g, HT42B534 have encountered this problem.A lot of boards are from Taobao.
- 0.3s is what I actually tested out.And this time interval is not in every package. Is the time of each command, I think it is acceptable.Smaller time will also be an error.

```python
def slip_reader(port):
    """Generator to read SLIP packets from a serial port.
    Yields one full SLIP packet at a time, raises exception on timeout or invalid data.

    Designed to avoid too many calls to serial.read(1), which can bog
    down on slow systems.
    """
    partial_packet = None
    in_escape = False
    while True:
        waiting = port.inWaiting()
        read_bytes = port.read(1 if waiting == 0 else waiting)
        print('read_bytes = ', read_bytes)
        if read_bytes == b'':
            raise FatalError("Timed out waiting for packet %s" % ("header" if partial_packet is None else "content"))
        for b in read_bytes:

            if type(b) is int:
                b = bytes([b])  # python 2/3 compat

            if partial_packet is None:  # waiting for packet header
                if b == b'\xc0':
                    partial_packet = b""
                else:
                    raise FatalError('Invalid head of packet (%r)' % b)
            elif in_escape:  # part-way through escape sequence
                in_escape = False
                if b == b'\xdc':
                    partial_packet += b'\xc0'
                elif b == b'\xdd':
                    partial_packet += b'\xdb'
                else:
                    raise FatalError('Invalid SLIP escape (%r%r)' % (b'\xdb', b))
            elif b == b'\xdb':  # start of escape sequence
                in_escape = True
            elif b == b'\xc0':  # end of packet
                yield partial_packet
                partial_packet = None
            else:  # normal byte in packet
                partial_packet += b
```
When I printed read_bytes in the slip_reader () function.Then try to read chip_id.
The normal chip is like this.
```
➜  esptool git:(master) ✗ python esptool.py -p /dev/tty.usbmodem0001 chip_id
esptool.py v2.2-dev
before connect function
Connecting...before enter self.read() function
read_bytes =  b'\xc0'
timeout =  0.1
read_bytes =  b'\x01\x08\x02\x00\x07\x07\x12 \x00\x00\xc0\xc0\x01\x08\x02\x00\x07\x07\x12 \x00\x00\xc0\xc0\x01\x08\x02\x00\x07\x07\x12 \x00\x00\xc0\xc0\x01\x08\x02\x00\x07\x07\x12 \x00\x00\xc0\xc0\x01\x08\x02\x00\x07\x07\x12 \x00\x00\xc0\xc0\x01\x08\x02\x00\x07\x07\x12 \x00\x00\xc0\xc0\x01\x08\x02\x00\x07\x07\x12 \x00\x00\xc0\xc0\x01\x08\x02\x00\x07\x07\x12 \x00\x00\xc0'
timeout =  0.1
after enter self.read() function
before enter self.read() function
after enter self.read() function
before enter self.read() function
after enter self.read() function
before enter self.read() function
after enter self.read() function
before enter self.read() function
after enter self.read() function
before enter self.read() function
after enter self.read() function
before enter self.read() function
after enter self.read() function
before enter self.read() function
after enter self.read() function

Detecting chip type...before enter self.read() function
read_bytes =  b'\xc0'
timeout =  3
read_bytes =  b'\x01\n\x02\x00\x00 \x06\x00\x00\x00\xc0'
```
Abnormal chip is like this.
```
before enter self.read() function
read_bytes =  b'\xc0'
timeout =  0.1
read_bytes =  b'\x01'
timeout =  0.1
read_bytes =  b'\x08'
timeout =  0.1
read_bytes =  b'\x02'
timeout =  0.1
read_bytes =  b'\x00\x07'
timeout =  0.1
read_bytes =  b'\x07'
timeout =  0.1
read_bytes =  b'\x12'
timeout =  0.1
read_bytes =  b' '
timeout =  0.1
read_bytes =  b'\x00'
timeout =  0.1
read_bytes =  b'\x00'
timeout =  0.1
read_bytes =  b'\xc0'
timeout =  0.1
after enter self.read() function

Detecting chip type...before enter self.read() function
read_bytes =  b'' 
timeout =  3

A fatal error occurred: Timed out waiting for packet header
```
The same before printing ``Detecting chip type...``.But not the same after printing.Abnormal chip can not receive data until the timeout.
I think the chip may not receive the order in time, or did not receive completely, and therefore can not respond.So I try to add a delay after the command has been sent and everything is OK.
However, what the specific problems are caused by, I am still not very understanding, hoping to be helpful.Thank you.

projectgus:
> And this time interval is not in every package. Is the time of each command, I think it is acceptable

Multiple calls to command() are made for each invocation of esptool.py, so delaying 300ms each time will have a major performance impact across all platforms and serial devices. I'd prefer to find a better way to fix it if we can.

I don't have a macOS computer I can try to reproduce this on, but to help us understand the bug can you please use dtruss to capture the OS-level sytem calls. Something like:

```
sudo dtruss esptool.py write_flash (args) 2> esptool_log.txt
```

Please first take a dtruss log without this PR so the bug is triggered, and then a log with this change applied where the bug is fixed.

Also, the PR you sent currently won't work due to an invalid name - please check the error shown above and fix the code.

projectgus:
Hi @rmokerone ,

I had an idea for a (possible) alternative fix for the serial driver issue, that won't reduce performance at all. Can you please try this branch and see if it changes at all:
https://github.com/espressif/esptool/tree/experiment/ch340_timeout_bug

Also, there is no a new option `esptool.py --trace`. If the branch doesn't work, can you please post the `esptool.py --trace` output when it fails. Thanks.

loadedtech:
I'm having problems with one of my devboards -- most work fine but after one flash it now doesn't work -- keep getting a timeout.

here's the trace using the branch above from the command ./esptool.py --port /dev/cu.wchusbserial1450 erase_flash

> TRACE +0.095 Read 1 bytes: b'\xc0'
> TRACE +0.000 Read 11 bytes: b'\x01\x07\x02\x00\x00\x00;\x86\x00\x00\xc0'
> TRACE +0.000 Full packet: b'\x01\x07\x02\x00\x00\x00;\x86\x00\x00'
> Running stub...
> TRACE +0.000 command op=0x06 data len=8 wait_response=1 timeout=3.000 data=b'\x00\x00\x00\x00\x04\x00\x10@'
> TRACE +0.000 Write 18 bytes: b'\xc0\x00\x06\x08\x00\x00\x00\x00\x00\x00\x00\x00\x00\x04\x00\x10@\xc0'
> TRACE +0.027 Read 1 bytes: b'\xc0'
> TRACE +0.000 Read 17 bytes: b'\x01\x06\x02\x00\x00\x00;\x86\x00\x00\xc0\xc0OHAI\xc0'
> TRACE +0.000 Full packet: b'\x01\x06\x02\x00\x00\x00;\x86\x00\x00'
> TRACE +0.000 Full packet: b'OHAI'
> Stub running...
> Erasing flash (this may take a while)...
> TRACE +0.000 command op=0xd0 data len=0 wait_response=1 timeout=120.000 data=b''
> TRACE +0.000 Write 10 bytes: b'\xc0\x00\xd0\x00\x00\x00\x00\x00\x00\xc0'
> TRACE +0.024 Read 1 bytes: b'\xc0'
> TRACE +0.000 Read 8 bytes: b'\x01\xd0\x00\x00\x00\x00\x00\x00'
> ^@TRACE +120.301 Timed out waiting for packet content
> 
> A fatal error occurred: Timed out waiting for packet content

Any tips appreciated

projectgus:
Hi @rmokerone ,

I think we've found the underlying issue causing this problem. Can you please try the latest master branch (with commit linked above) and let me know if it fixes the problem.

Angus

dehart:
I've just tried master on MacOS got same result as @loadedtech
`python esptool.py --trace --baud 115200 -p /dev/cu.wchusbserial1420 erase_flash`
Output
```
......
TRACE +0.000 Full packet: 'OHAI'
Stub running...
Erasing flash (this may take a while)...
TRACE +0.000 command op=0xd0 data len=0 wait_response=1 timeout=120.000 data=''
TRACE +0.000 Write 10 bytes: '\xc0\x00\xd0\x00\x00\x00\x00\x00\x00\xc0'
TRACE +0.024 Read 1 bytes: '\xc0'
TRACE +0.000 Read 8 bytes: '\x01\xd0\x00\x00\x00\x00\x00\x00'
TRACE +120.007 Timed out waiting for packet content

A fatal error occurred: Timed out waiting for packet content
```
Led is no longer blinking when at "Read 8 bytes".
Not sure if this is normal.

projectgus:
@loadedtech @dehart I don't think this is the same issue addressed in this PR. The timeout is happening much later, and it doesn't fit the "failing to read after write" driver bug that @rmokerone is mentioning. I've opened issue #1848 to track this, please take a look there.

@rmokerone I'm going to close this as I haven't heard back from you, but if you get a chance to look at this again then please let me know. Thanks.

PR #281, opened 2018-03-05
Change shebang reference to 'python' to be 'python2' instead

SteveOfTheStow:
# Description of change

PEP-394 recommends that scripts only use the 'python' command in the shebang if they are source compatible with both Python2 and Python3. Over time environments will migrate towards pointing 'python' at 'python3', meaning scripts that are only python2 compatible will likely fail. esptool is included in this list.

Most recently, the popular Homebrew package manager for macOS made this change: https://brew.sh/2018/01/19/homebrew-1.5.0/, which broke my workflow, so thought it best to help everyone out.

There is no effect for those whose python command still points to python2.

macOS 10.13, ESP-WROVER-KIT, ESP32

./test_imagegen.py
.........
Ran 9 tests in 1.719s

OK

./test_esptool.py /dev/tty.usbserial-14101 esp32 230400
Running esptool.py tests...
........................................
Ran 40 tests in 442.823s

OK

### DIFF
--- a/ecdsa/ecdsa.py
+++ b/ecdsa/ecdsa.py
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
+#! /usr/bin/env python2
 
 """
 Implementation of Elliptic-Curve Digital Signatures.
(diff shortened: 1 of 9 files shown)


SteveOfTheStow:
Just read that esptool is meant to be source-compatible with 2 and 3!

PR #282, opened 2018-03-07
Update README.md

sajink:
Added a line about windows driver in the readme.md file. No code changes involved.


projectgus:
Hi @sajink ,

Thanks for thinking of this point, I agree it's a good place to mention the need for drivers.

I'm not sure about linking directly to SiLabs, because there are at least 4-5 different USB/serial vendors who make drivers for common chipsets. Maybe the README could suggest users check the Windows Device Manager for any devices with missing drivers, or something like this? 

Angus

PR #285, opened 2018-03-23
esptool: Add new argument "--no-sync" to skip sync and initial baud

tim-nordell-nimbelink:
We have a board designed which has the UART to the ESP32 as a passthrough from
another microcontroller.  As such, it's easier to send the sync sequence and
change the baud rate on the microcontroller side of things than to pass this
change through from the PC.  Adding the --no-sync option allows the intermediate
microcontroller to run the sync sequence and to set the appropriate baud rate
on the ESP32 side of things.  For this use case, we need to avoid sending the
change buad rate sequence (the PC baud rate is not necessarily the same as
the microcontroller <-> ESP32) and avoid sending the sync sequence.

Signed-off-by: Tim Nordell <tim.nordell@nimbelink.com>

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -365,6 +365,11 @@ class ESPLoader(object):
         # Details: https://github.com/espressif/esptool/issues/136
         last_error = None
 
+        # If we're doing no_sync, we're likely communicating as a pass through
+        # with an intermediate device to the ESP32
+        if mode == "no_reset_no_sync":
+            return last_error
+
         # issue reset-to-bootloader:
         # RTS = either CH_PD/EN or nRESET (both active low = chip in reset
         # DTR = GPIO0 (active low = boot to flasher)
@@ -2158,7 +2163,7 @@ def main():
     parser.add_argument(
         '--before',
         help='What to do before connecting to the chip',
-        choices=['default_reset', 'no_reset'],
+        choices=['default_reset', 'no_reset', 'no_reset_no_sync'],
         default=os.environ.get('ESPTOOL_BEFORE', 'default_reset'))
 
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


esptool.py
```diff
@@ -380,34 +380,35 @@ def _connect_attempt(self, mode='default_reset', esp32r0_delay=False):
...
-            except FatalError as e:
-                if esp32r0_delay:
-                    print('_', end='')
-                else:
-                    print('.', end='')
-                sys.stdout.flush()
-                time.sleep(0.05)
-                last_error = e
-        return last_error
+        if not no_sync:
```
projectgus: Rather than indenting all the sync logic, how about returning here if `no_sync` is set?

tim-nordell-nimbelink: Noted.

tim-nordell-nimbelink:
Would the --baud setting for the espefuse.py be confusing for users that don't have the middleman pass through?

projectgus:
Hi @tim-nordell-nimbelink ,

This looks good. Adding `--baud` to espefuse.py seems fine, there are other conceivable (if rare) reasons someone may find a use for this, such as having a serial link which is unreliable at high baud rates.

Do you need `--before no_sync_no_reset` to work in espefuse.py as well? If so, it might be time to refactor some of these common arguments out into a utility function in esptool.py, for espefuse to import.

One minor nitpack, I think I prefer `no_reset_no_sync` to `no_reset`, as the order of operations is reset-then-sync.

The only other thing is the CI PEP8 failures, from running the [flake8 tool](http://flake8.readthedocs.org/). You can click on the link above to see details.

tim-nordell-nimbelink:
Hi @projectgus ,

No, the no_sync_no_reset for espefuse.py isn't needed - not sure why and I haven't dug into it.  I'll add it in for consistency though.  (Maybe because it wasn't trying to change the baud rate?)

- Tim

projectgus:
Hi @tim-nordell-nimbelink ,

Sorry, I missed that you'd already made the change to espefuse.py in the commits before you left the comment. Will merge now.

Angus

tim-nordell-nimbelink:
Hi @projectgus ,

No worries!  It just means I can use the vanilla repository now instead of my own modifications locally. :)

 -- Tim

Issue #190, opened 2017-03-29
i can not receive RTS  info


ouki-wang:
I used a standard USB CDC device（stm32）
when i close the serial，i receive the RTS info
21 22 02 00 00 00 00 00

but when I use esptool.exe，I can not receive 21 22 02 00 00 00 00 00 to control RTS pin

ouki-wang:
can you enable RTS control？

ouki-wang:
will esptool.exe source code be opened?

projectgus:
What esptool.exe are you referring to?

ouki-wang:
@projectgus run get.exe. I can get a esptool.exe

ouki-wang:
I have found a project source https://github.com/igrr/esptool-ck
Is is an official release version?

projectgus:
I'm sorry but I don't understand the problem you are having at all, I need more details.

* Are you using Arduino for ESP32 to install esptool.exe?
* What device is connected to the STM32 CDC firmware?
* Where does the CDC firmware source code from?
* What host OS are you on?
* Can you post the output from a failed flash with esptool?

projectgus:
Closing due  to a lack of information to follow up, but feel free to reopen if you have more information.

PR #290, opened 2018-03-28
Add work-around for RTS control for CDC-ACM serial adapters using usbser.sys

devanlai:
## Description of change
The usbser.sys driver used for standard CDC-ACM serial adapters on Windows has a [known limitation](https://answers.microsoft.com/en-us/windows/forum/windows_xp-hardware/virtual-serial-port-usbsersys-is-not-sending/a086cbf6-2d0a-4c46-88fd-73865bc185f5) where it will only send the set-control-line-state request when DTR is modified instead of when either DTR or RTS are modified.

As a work-around, send no-op DTR writes after RTS writes so the driver will send the set-control-line-state request with the updated RTS state.

Prior to this work-around, CDC-ACM adapters controlled from Windows would effectively generate the following DTR/RTS sequence for resets:

- DTR -> High
- sleep 0.1
- DTR -> Low, RTS -> Low
- sleep 0.05
- DTR -> High, RTS -> High

## This change fixes the following bug(s):
#190 

## I have tested this change with the following hardware & software combinations:
Windows 10, custom PCB and custom CDC-ACM usb-serial adapter at 115200 baud, ESP8266 on an ESP-WROOM-02 module.

## I have run the esptool.py automated integration tests with this change and the above hardware. The results were:

    >python test\test_esptool.py COM8 esp8266 115200
    Running esptool.py tests...
    ........................................
    Ran 40 tests in 675.891s
    
    OK

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -342,6 +342,16 @@ class ESPLoader(object):
         for i in range(7):
             self.command()
 
+    def _setDTR(self, state):
+        self._port.setDTR(state)
+
+    def _setRTS(self, state):
+        self._port.setRTS(state)
+        # Work-around for adapters on Windows using the usbser.sys driver:
+        # generate a dummy change to DTR so that the set-control-line-state
+        # request is sent with the updated RTS state and the same DTR state
+        self._port.setDTR(self._port.dtr)
+
     def _connect_attempt(self, mode='default_reset', esp32r0_delay=False):
         """ A single connection attempt, with esp32r0 workaround options """
         # esp32r0_delay is a workaround for bugs with the most common auto reset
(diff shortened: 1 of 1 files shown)


projectgus:
Hi @devanlai ,

Thanks for sending this. I ran the test suite on Linux and it also runs fine (I can't think of a way this would break something, but as we've seen here hardware driver behaviour can be deceptive...)

BTW, if you're feeling enthusiastic then it would also be nice to add this workaround upstream in pyserial, it's probably the more appropriate place. Looks like a pretty straightforward change, here:
https://github.com/pyserial/pyserial/blob/master/serial/serialwin32.py#L373

But I'm happy to merge this now. If a fix goes upstream and becomes available in a release of pyserial, we can roll this back then.

Angus

devanlai:
Thanks for merging this.

For fixing it upstream, I'll do some research to see if it's possible to reliably determine when the serial port driver is using usbser.sys so that the workaround is applied as narrowly as possible.

PR #307, opened 2018-05-14
Improve usability (default COM port) when programming

Callahan93:
# Description of change

### Added 'Auto COM Port' as the default action when no port is specified.
The default COM port is `DEFAULT_PORT = "/dev/ttyUSB0"` which is incorrect for Windows machines and some users (less technical) have difficulty finding the COM port (often it is a dual port).
If no port is parsed in the CLI, then all available ports are discovered and attempted to connect the ESP device. The first successful one is used thereafter. If the port is passed through the CLI then the normal operation takes place.

### Reduced the number of connection attempts from 10 to 3 
Auto COM finder could take ages if it's 10, also if no connection after 3 attempts its unlikely to succeed.

### Print the MAC address when programming

Windows 7, Python3.6, ESP32-WROOM

Results:
`esptool.py v2.4.0-dev
Port not specified!      Using port auto find...
['COM1', 'COM15', 'COM16']
Trying COM1
Connecting........_____....._____....._____
Trying COM15
Connecting........_____....._____....._____
Trying COM16
Connecting.......
Chip is ESP32D0WDQ6 (revision 1)
Features: WiFi, BT, Dual Core`

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -2218,6 +2218,7 @@ def version(args):
 # End of operations functions
 #
 
+
 def all_serial_ports():
     import serial.tools.list_ports as list_ports
 
--- a/esptool.py
+++ b/esptool.py
@@ -31,6 +31,7 @@ import sys
 import time
 import zlib
 import string
+import serial.tools.list_ports as list_ports
 
 import serial
 
@@ -2219,16 +2220,6 @@ def version(args):
 #
 
 
-def all_serial_ports():
-    import serial.tools.list_ports as list_ports
-
-    result = []
-    for port in list_ports.comports():
-        result.append(port.device)
-    result.sort()
-    return result
-
-
 def main():
     parser = argparse.ArgumentParser(description='esptool.py v%s - ESP8266 ROM Bootloader Utility' % __version__, prog='esptool')
 
@@ -2448,10 +2439,8 @@ def main():
 
     if operation_args[0] == 'esp':  # operation function takes an ESPLoader connection object
         initial_baud = min(ESPLoader.ESP_ROM_BAUD, args.baud)  # don't sync faster than the default baud rate
-        ser_list = all_serial_ports() if args.port is None else [args.port]
-        ser_attempts = 0
+        ser_list = sorted(ports.device for ports in list_ports.comports())
         for each_port in reversed(ser_list):
-            ser_attempts += 1
             if args.port is None:
                 print("Trying %s ... " % each_port)
             try:
--- a/esptool.py
+++ b/esptool.py
@@ -2214,34 +2214,7 @@ def version(args):
 # End of operations functions
 #
 
-
 def all_serial_ports():
-    """
-    https://stackoverflow.com/questions/12090503/listing-available-com-ports-with-python
-    """
-    import glob
-
-    if sys.platform.startswith('win'):
-        ports = ['COM%s' % (i + 1) for i in range(256)]
-    elif sys.platform.startswith('linux') or sys.platform.startswith('cygwin'):
-        # This excludes your current terminal "/dev/tty"
-        ports = glob.glob('/dev/tty[A-Za-z]*')
-    elif sys.platform.startswith('darwin'):
-        ports = glob.glob('/dev/tty.*')
-    else:
-        raise EnvironmentError('Unsupported platform')
-    result = []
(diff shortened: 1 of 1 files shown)


esptool.py
```diff
@@ -2214,6 +2217,35 @@ def version(args):
 #
 
 
+def all_serial_ports():
```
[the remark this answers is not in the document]
Callahan93: Done `all_serial_ports_2` with the PySerial method. See https://stackoverflow.com/questions/12090503/listing-available-com-ports-with-python
that original method may be more reliable? 
Both work fine on my machine. What do you prefer?

esptool.py
```diff
@@ -2433,20 +2465,47 @@ def add_spi_flash_subparsers(parent, is_elf2image):
...
+                        continue
+                break
         else:
-            chip_class = {
-                'esp8266': ESP8266ROM,
-                'esp32': ESP32ROM,
-            }[args.chip]
-            esp = chip_class(args.port, initial_baud, args.trace)
-            esp.connect(args.before)
+            if args.chip == 'auto':
```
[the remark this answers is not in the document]
Callahan93: Done

esptool.py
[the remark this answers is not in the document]
projectgus: Hi @Callahan93 ,

Thanks for being patient while I got back to you.

I prefer the pyserial method. The code at the linked StackOverflow issue works fine for me (Windows & Linux), so I suspect any underlying bug in pyserial is long fixed. If it's possible not to reinvent a wheel, then better not to.

esptool.py
```diff
@@ -2433,15 +2470,28 @@ def add_spi_flash_subparsers(parent, is_elf2image):
...
+                else:
+                    chip_class = {
+                        'esp8266': ESP8266ROM,
+                        'esp32': ESP32ROM,
+                    }[args.chip]
+                    esp = chip_class(each_port, initial_baud, args.trace)
+                    esp.connect(args.before)
+                break
+            except FatalError as err:
+                if ser_attempts >= len(ser_list) or args.port is not None:
```
projectgus: This is a bit of an awkward way to check if you're out of options, and also it swallows the exception object details (stack trace, etc) for a failed connection (which may make debugging harder for a specified port). How about something like:

```
except FatalError as err:
   if args.port is not None:
       raise
   print("%s failed to connect: %s" % (each_port, err))
   esp = None

then, outside the loop...

```
if esp is None:
   raise FatalError("None of %d available serial ports could connect." % len(ser_list))
```

File: esptool.py [hunk 4]
```diff
@@ -2213,6 +2214,15 @@ def version(args):
 # End of operations functions
 #
 
+def all_serial_ports():
```
projectgus: I think this function can be replaced with `list(sorted(list_ports.comports()))` ) or possibly just `sorted(list_ports.comports())`.

Please move the `import` to the import block at the top of the file.

esptool.py
```diff
@@ -2433,15 +2443,30 @@ def add_spi_flash_subparsers(parent, is_elf2image):
...
-            chip_class = {
-                'esp8266': ESP8266ROM,
-                'esp32': ESP32ROM,
-            }[args.chip]
-            esp = chip_class(args.port, initial_baud, args.trace)
-            esp.connect(args.before)
+        ser_list = all_serial_ports() if args.port is None else [args.port]
+        ser_attempts = 0
+        for each_port in reversed(ser_list):
+            ser_attempts += 1
```
projectgus: This variable is no longer used for anything

esptool.py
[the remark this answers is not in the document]
Callahan93: No problem, all done

esptool.py
Callahan93: Sorted

esptool.py
Callahan93: Will try and look at this today

Callahan93: Put into a single line instead of a function

esptool.py
Callahan93: done

projectgus:
Hi @Callahan93 ,

Thanks for the additional changes. I rebased your changes on master (to resolve the conflict) and squashed commits. The result has been cherry-picked as 6575a49.

Angus

PR #309, opened 2018-05-17
Print MAC before programming

Callahan93:
# Description of change

Print the MAC address when programming - This is useful when you need to keep track of multiple modules being flashed.

Windows 7, Python3.6, ESP32-WROOM


projectgus:
Good idea, thanks!

Issue #313, opened 2018-06-01
Serial port not closed explicitly in main()


cefn:
(Please delete any lines which don't apply.)

* Operating system:

All, but particularly affects Windows because of the implementation of pyserial.

* Python version: (`python -V` to check this)

All

* ESP hardware in use:

All, but demonstrated with NodeMCU-M ESP8285 running over CH340

# Full esptool.py command line as run:

```python
import sys
import esptool

sys.argv = "esptool.py --after hard_reset --port COM3 flash_id".split()
esptool.main()
sys.argv = "esptool.py --after hard_reset --port COM3 erase_flash".split()
esptool.main()
sys.argv = "esptool.py --after hard_reset --port COM3 --baud 1500000 write_flash --flash_mode dout --flash_size 1MB 0 C:\\Users\\cefn\\Desktop\\vgkits-vanguard\\vgkits\\vanguard\\data\\firmware\\vanguard-0.1.0.bin".split()
esptool.main()
```

# Full output from esptool.py (please copy and paste all lines of output)

[...]

PR #316, opened 2018-06-04
Close serial port before exit from main()

cefn:
Addresses the issue detailed at https://github.com/espressif/esptool/issues/313 and has the close call inside the proper if block.


cefn:
OK, this seems like the flake8 checks have passed and it also has the proper history. Hope it is OK.

PR #326, opened 2018-07-06
Set serial port write timeout in esptool

heyinling:
I'm using esptool get chip id function to confirm if the listed port is a valid ESP port. One issue I met is Raspberry Pi list port result contains `/dev/ttyAMA0`. Writing to this port will never return, as we don't set write timeout.

Operating system: MAC OSX, Raspbian
development board name: ESP32 DevKitC

(This case also fails on my side even use master)
```
Running esptool.py tests...
................................F
Stdout:
**************************************************
Running /usr/local/opt/python/bin/python2.7 /Users/heyinling/workspace/esptool/test/../esptool.py --chip esp32 --port /dev/tty.SLAB_USBtoUART --baud 230400 load_ram images/helloworld-esp32.bin...
esptool.py v2.4.1
Serial port /dev/tty.SLAB_USBtoUART
Connecting........__
Chip is ESP32D0WDQ6 (revision 0)
Features: WiFi, BT, Dual Core
MAC: 24:0a:c4:04:26:5c
Uploading stub...
Running stub...
Stub running...
Changing baud rate to 230400
Changed.
RAM boot...
Downloading 124 bytes at 4008c000... done!
All segments done, executing at 4008c004
Exiting immediately.

........
======================================================================
FAIL: test_load_ram (__main__.TestLoadRAM)
Verify load_ram command
  File "./test/test_esptool.py", line 396, in test_load_ram
AssertionError: 'Hello world!' not found in '\x1e\xb7|\x1e\xf5\x1e?\xf4T\xf5\\?~\\\xb7|\\?\xf4t\xf5\x1e\xb7|\\\xb7|t\x7ft\x1e\xb7'

Stdout:
**************************************************
Running /usr/local/opt/python/bin/python2.7 /Users/heyinling/workspace/esptool/test/../esptool.py --chip esp32 --port /dev/tty.SLAB_USBtoUART --baud 230400 load_ram images/helloworld-esp32.bin...
esptool.py v2.4.1
Serial port /dev/tty.SLAB_USBtoUART
Connecting........__
Chip is ESP32D0WDQ6 (revision 0)
Features: WiFi, BT, Dual Core
MAC: 24:0a:c4:04:26:5c
Uploading stub...
Running stub...
Stub running...
Changing baud rate to 230400
Changed.
RAM boot...
Downloading 124 bytes at 4008c000... done!
All segments done, executing at 4008c004
Exiting immediately.


Ran 41 tests in 676.930s

FAILED (failures=1)
```


esptool.py
```diff
@@ -294,6 +294,7 @@ def command(self, op=None, data=b"", chk=0, wait_response=True, timeout=DEFAULT_
         new_timeout = min(timeout, MAX_TIMEOUT)
         if new_timeout != saved_timeout:
             self._port.timeout = new_timeout
+            self._port.write_timeout = new_timeout
```
projectgus: Setting the same as the command timeout means that some commands (like erase) will have very long write timeouts (120 seconds).

I think it's probably safe to set write_timeout when the port is opened and not change it again. The most data esptool sends in a single command is 16KB (writing flash to ESP32). This can take a long time at a low baud rate, but looking at the pyserial code it looks like the write timeout is only until the OS buffers the write, and I think all OSes will buffer more than 16KB. So the write command itself should always return almost instantly, if the port is available and working correctly. Maybe the write timeout could be something like 3 seconds, to give some headroom?

heyinling: I think we can set a long timeout, 30 or 60 seconds for example, when opening the port. It will not affect normal use cases, only make sure esptool will finally exit in the rare scenario I mentioned in commit message.

projectgus: Because we have the "scan" feature if no port is given, it's useful to have esptool move on to a new port if it can't write to one for some reason. Maybe 10 seconds?

heyinling: Sure. I have changed to 10 seconds.

projectgus:
Thanks @heyinling !

PR #344, opened 2018-08-06
add argument to erase flash before flashing

Callahan93:
# Description of change

Sometimes we want to erase the whole chip because we use some partitions for other data (like OTA) which is not overwritten when flashed with new firmware. Currently, this makes this a 2 stage process (erase_flash ... then write_flash) - This change is intended to make this a one stage command.

As this is a new argument please advise if you prefer another format.

ESP32, Windows 7, Python 3.6

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -2048,6 +2048,9 @@ def write_flash(esp, args):
                              % (argfile.name, argfile.tell(), address, flash_end))
         argfile.seek(0)
 
+    if args.erase_all:
+        erase_flash(esp, args)
+
     for address, argfile in args.addr_filename:
         if args.no_stub:
             print('Erasing flash...')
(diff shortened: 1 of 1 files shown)


marcelstoer:
Since [v2.6](https://github.com/espressif/esptool/releases/tag/v2.6) is out now this could be closed I guess.

projectgus:
@marcelstoer You're too quick for me. It can indeed, cherry-picked as 822112b (small tweak: used a hyphen instead of an underscore for `--erase-all`. Eventually I want to standardize on hyphens for all commands and options, probably as a breaking change in v3.0).

There are probably some optimisations possible with this option, at least with the stub, because sending "Sector Erase"/"Block Erase" commands are redundant if we know the whole chip is already erased. But this is fine for now (and sector erases take less time when the chip is erased, so the performance hit is not as bad as it could be!).

PR #386, opened 2018-12-13
esptool: add raw flashing mode to avoid breaking encrypted images

grandcat:
# Description of change
## Current state 
Before flashing, the `esptool` modifies the header (first 4 bytes) of the bootloader image. Bootloader image is recognized by comparing a single magic byte.
## Problem statement
By flashing images previously encrypted with `espsecure`, there is a chance of `P > 1/256` that an encrypted image is falsely assumed to be a bootloader image.
This will cause the `esptool` to overwrite the first 4 bytes in any case, thus corrupting the encrypted image.
## Consequence
Depending on the chosen key and the image, there is a chance that `esptool` silently modifies the image and thus, it won't boot. 

For flashing and flash verification, a `--raw-mode` or `-r` argument was introduced. 
It basically makes the `esptool` flash the image as it is without any attempt for modifications.

The issue was previously discussed with our contact as Espressif.

Ubuntu 18.04 and ESP32 (4mb and 8mb variant)

```
Running esptool.py tests...
.........................................
Ran 41 tests in 656.201s

OK
```

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -2057,7 +2057,14 @@ def detect_flash_size(esp, args):
 
 
 def _update_image_flash_params(esp, address, args, image):
-    """ Modify the flash mode & size bytes if this looks like an executable bootloader image  """
+    """
+    Modify the flash mode & size bytes if this looks like an executable bootloader image.
+
+    Avoid changing any bytes in case the image is encrypted as the key is unknown to this tool.
+    """
+    if args.raw_mode:
+        return image  # Do not touch the image. It might be encrypted.
+
     if len(image) < 8:
         return image  # not long enough to be a bootloader image
 
@@ -2466,6 +2473,8 @@ def main():
     parser_write_flash.add_argument('addr_filename', metavar='<address> <filename>', help='Address followed by binary filename, separated by space',
                                     action=AddrFilenamePairAction)
     add_spi_flash_subparsers(parser_write_flash, is_elf2image=False)
+    parser_write_flash.add_argument('--raw-mode', '-r', help='Do not overwrite flash properties in image header. Must be set if image is encrypted.',
+                                    action="store_true")
     parser_write_flash.add_argument('--no-progress', '-p', help='Suppress progress output', action="store_true")
     parser_write_flash.add_argument('--verify', help='Verify just-written data on flash ' +
                                     '(mostly superfluous, data is read back during flashing)', action='store_true')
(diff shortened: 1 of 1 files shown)


projectgus:
Hi @grandcat ,

Thanks for sending the patch for this. I think I also sent some earlier comments on this approach via email, which hopefully got back to you.

Rather than adding a new option for this, I'd like to improve the heuristic for "is this an image?" by using esptool's existing capability to parse an ESP32 binary image and verify it (using CRC and optionally an appended SHA hash). If there's an error in the binary image when parsed this way, it can be treated it in the way that "raw" mode works in this PR. This should remove false-positives for encrypted binaries that happen to start with the "magic" byte.

The other thing that can be done is to add a `--flash-size keep` value to write_flash. There are already `--flash-mode keep` and `--flash-speed keep` options (which are the defaults), and if all these options are set then the header also won't be touched during flashing.

If you're interested in working on either of these approaches then that would be great. Otherwise I can probably make the changes soon.

grandcat:
Thanks for your input and sorry for the late reply.
I will have a look at the additional `--flash-size keep` option. However, it obfuscates the issue in this tool imo.

For your first approach about parsing the ESP32 binary image. This would mean that the flash tool is aware of the encryption key, isn't it? Otherwise, the byte sequence is still just random.

projectgus:
@grandcat I'm sorry that "next couple of weeks" turned into 6 months! :grimacing: 

Please let me know if this alternative fix doesn't work for your purposes, for any reason.

grandcat:
Thanks @projectgus . 
I reviewed your fix, but need to verify that verification by
```
test_image = esp.BOOTLOADER_IMAGE(io.BytesIO(
test_image.verify()
```
is safe for encrypted images. I guess there is some checksum calculation (without having looked into it), so it should be fine.

Thanks for fixing, anyways :+1:

projectgus:
> is safe for all kind of encrypted images

@grandcat If fed ciphertext, this should throw some exception and fail (there is some extremely unlikely possibility that the ciphertext will have valid-seeming values and ranges for headers for a short valid binary, and then a valid checksum in the right place, but it's going to be a significantly smaller chance than before.)

To remove any doubt, pass the (also new in v2.7) `write_flash --flash-size keep` argument to esptool.py when flashing an image that you know is encrypted.

grandcat:
Ok, thank you very much for the explanation and the hint `keep`. This makes us confident.

PR #400, opened 2019-01-26
Ensure expand_file_arguments gets called when using custom_commandline (ESPTOOL-166)

eykamp:
Fix bug where expand_file_arguments is not called when using custom_commandline arguments. (Technically, it is called, it just doesn't do anything useful.)

Also makes the flow of arguments through the program slightly clearer by always presenting the authoritative arg list to main().

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -2608,7 +2608,7 @@ def main(custom_commandline):
 
     custom_commandline = expand_file_arguments(custom_commandline)
 
-    args = parser.parse_args()
+    args = parser.parse_args(custom_commandline)
 
     print('esptool.py v%s' % __version__)
 
--- a/esptool.py
+++ b/esptool.py
@@ -2392,7 +2392,7 @@ def version(args):
 #
 
 
-def main(custom_commandline=None):
+def main(custom_commandline):
     """
     Main function for esptool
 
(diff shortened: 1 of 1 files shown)


esptool.py
```diff
@@ -2392,7 +2392,7 @@ def version(args):
 #
 
 
-def main(custom_commandline=None):
+def main(custom_commandline):
```
projectgus: I've seen there is existing code out there which writes `sys.argv` and then calls `esptool.main()` (this behaviour predates the `custom_commandline` argument).

To avoid breaking this code, I'd prefer if the default argument was kept and somewhere in main() it does:

```
if custom_commandline is None:
   custom_commandline = sys.argv[1:]
```

(`sys.argv[1:]` can't be the default argument because it needs to be expanded when the function is called, not when esptool is imported).

If you're feeling super generous, it may be a good improvement to rename `custom_commandline` to something like `argv`, if it's going to be passed around in multiple places.

projectgus:
@eykamp I'm guessing you've moved on to other things, should I close these PRs?

eykamp:
I don't remember why this got dropped; maybe thought you were going to make the modification to the PR that you suggested, and you thought I was?  :-)

I'm guessing the underlying bug is still there.  I will probably be coming back to the larger project that led me to make the PR, but even if I don't I think it would make sense to fix.   You made two suggestions: checking if anything was passed with custom_commandline, and a suggestion for renaming.  I think both make sense.

I'm willing to follow your lead.  If you want to fix the problem, and want me to make the changes suggested, I will update the PR.  If you don't think this is an issue, we can drop it.  Either way I agree we should get the PR closed and off the books.

PR #401, opened 2019-01-27
Improve awkward error message

eykamp:
# Description of change
Improve an awkward error message -- no change in functionality


projectgus:
Thanks for the contribution and being patient while someone got to it, @eykamp  :)

Issue #406, opened 2019-02-10
args passed to main(...) are not parsed correctly


marcelstoer:
* Operating system: macOS
* Python version: (`python -V` to check this) 3.6.5
* ESP hardware in use: any

# What is the expected behaviour?
CLI args can be passed as-is to `esptool.main(...)` to get exactly the same behavior.

# Actual behaviour
All of the below combinations trip up the parser. It fails with (highlight is mine)

> usage: esptool [-h] [--chip {auto,esp8266,esp32}] [--port PORT] [--baud BAUD]
               [--before {default_reset,no_reset,no_reset_no_sync}]
               [--after {hard_reset,soft_reset,no_reset}] [--no-stub]
               [--trace] [--override-vddsdio [{1.8V,1.9V,OFF}]]
               {load_ram,dump_mem,read_mem,write_mem,write_flash,run,image_info,make_image,elf2image,read_mac,chip_id,flash_id,read_flash_status,write_flash_status,read_flash,verify_flash,erase_flash,erase_region,version}
               ...
esptool: error: argument operation: **invalid choice: '-'** (choose from 'load_ram', 'dump_mem', 'read_mem', 'write_mem', 'write_flash', 'run', 'image_info', 'make_image', 'elf2image', 'read_mac', 'chip_id', 'flash_id', 'read_flash_status', 'write_flash_status', 'read_flash', 'verify_flash', 'erase_flash', 'erase_region', 'version')

```
--port /dev/cu.SLAB_USBtoUART --baud 921600 --after no_reset write_flash --erase-all --flash_mode dio 0x00000 /Users/marcelstoer/Downloads/nodemcu-dev-17-modules-2019-02-10-13-23-19-float.bin
-p /dev/cu.SLAB_USBtoUART -b 921600 -a no_reset write_flash -e -fm dio 0x00000 /Users/marcelstoer/Downloads/nodemcu-dev-17-modules-2019-02-10-13-23-19-float.bin
--port /dev/cu.SLAB_USBtoUART --baud 921600 --after no_reset write_flash --flash_mode dio 0x00000 /Users/marcelstoer/Downloads/nodemcu-dev-17-modules-2019-02-10-13-23-19-float.bin --erase-all
--port /dev/cu.SLAB_USBtoUART --baud 921600 --after no_reset write_flash --flash_mode dio 0x00000 /Users/marcelstoer/Downloads/nodemcu-dev-17-modules-2019-02-10-13-23-19-float.bin
```

[...]

PR #408, opened 2019-02-13
Document passing arguments to main()

marcelstoer:
Addresses #406

### DIFF
--- a/README.md
+++ b/README.md
@@ -416,6 +416,19 @@ Note that not every serial program supports the unusual ESP8266 74880bps "boot l
 
 Running `esptool.py --trace` will dump all serial interactions to the standard output (this is *a lot* of output). This can be helpful when debugging issues with the serial connection, or when providing information for bug reports.
 
+## Using esptool.py from Python
+
+esptool.py can easily be integrated into Python applications or called from other Python scripts.
+
+While it currently does have a poor Python API, something which [#208](https://github.com/espressif/esptool/issues/208) will address, it allows for passing CLI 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -416,6 +416,19 @@ Note that not every serial program supports the unusual ESP8266 74880bps "boot l
...
+
+esptool.py can easily be integrated into Python applications or called from other Python scripts.
+
+While it currently does have a poor Python API, something which [#208](https://github.com/espressif/esptool/issues/208) will address, it allows for passing CLI 
+arguments to `esptool.main()`. This workaround makes integration very straightforward as you can pass exactly the 
+same arguments as you would on the CLI.
+
+```python
+command = '--baud 460800 read_flash 0 0x200000 flash_contents.bin'
+esptool.main(command.split(' '))
```
projectgus: `split(' ')` will work until there's a space in one of the pathnames :)

What do you think about `command = [ '--baud', '460800', 'read_flash', '0', '0x200000', 'flash_contents.bin']`

It's less readable but it doesn't contain any traps for the unsuspecting. :)

marcelstoer:
> Most likely this approach will remain compatible for all future versions though

It's a relief to hear that! For my use case that "API" is perfect. However, I do understand that any other application that integrates esptool to use only individual functions needs something else.

README.md
marcelstoer: > `split(' ')` will work until there's a space in one of the pathnames :)

Argghh, you're right, thanks. I hardly ever think of that because using spaces feels like such a weird thing to do 😉

PR #416, opened 2019-03-04
Improve encryption speed

darkfader:
# Description of change
[python2+3] Increase speed of encryption using pycrypto if it is installed.
[python3] Instead of going through 256 bits for key tweaking, go through the address offset bits only. Also in combination with bigint.

macOS, Python3.7, ESP32, pre-programmed encryption key.
flash_crypt_config=0xF but also compared outputs on 0x1 and 0x2.

### DIFF
--- a/espsecure.py
+++ b/espsecure.py
@@ -254,6 +254,11 @@ _FLASH_ENCRYPTION_TWEAK_PATTERN = [
 ]
 assert len(_FLASH_ENCRYPTION_TWEAK_PATTERN) == 256
 
+_FLASH_ENCRYPTION_TWEAK_PATTERN_REVERSE = [0] * 24
+
+for i in range(0, 256):
+    address_bit = _FLASH_ENCRYPTION_TWEAK_PATTERN[i]
+    _FLASH_ENCRYPTION_TWEAK_PATTERN_REVERSE[address_bit] ^= 1 << (255 - i)
 
 def _flash_encryption_tweak_range(flash_crypt_config=0xF):
     """ Return a list of the bit indexes that the "key tweak" applies to,
@@ -271,6 +276,21 @@ def _flash_encryption_tweak_range(flash_crypt_config=0xF):
     return tweak_range
 
 
+def _flash_encryption_tweak_range_bits(flash_crypt_config=0xF):
+    """ Return a bits that the "key tweak" applies to,
+    as determined by the FLASH_CRYPT_CONFIG 4 bit efuse value.
+    """
+    tweak_range = 0
+    if (flash_crypt_config & 1) != 0:
+        tweak_range |= 0xFFFFFFFFFFFFFFF8000000000000000000000000000000000000000000000000
+    if (flash_crypt_config & 2) != 0:
+        tweak_range |= 0x0000000000000007FFFFFFFFFFFFFFF000000000000000000000000000000000
+    if (flash_crypt_config & 4) != 0:
+        tweak_range |= 0x0000000000000000000000000000000FFFFFFFFFFFFFFFF80000000000000000
+    if (flash_crypt_config & 8) != 0:
+        tweak_range |= 0x000000000000000000000000000000000000000000000007FFFFFFFFFFFFFFFF
+    return tweak_range
+
 def _flash_encryption_tweak_key(key, offset, tweak_range):
     """Apply XOR "tweak" values to the key, derived from flash offset
     'offset'. This matches the ESP32 hardware flash encryption.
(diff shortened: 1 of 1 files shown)


espsecure.py
```diff
@@ -274,22 +307,27 @@ def _flash_encryption_tweak_key(key, offset, tweak_range):
...
+                key[bit // 8] ^= 1 << (7 - (bit % 8))
+
+        key = b"".join(chr(k) for k in key)
+        return key
 
-    if esptool.PYTHON2:
-        return b"".join(chr(k) for k in key)
     else:
-        return bytes(key)
+        key = int.from_bytes(key, byteorder='big', signed=False)
```
projectgus: This looks like it's probably relatively expensive. Is there a way to take it out of the loop?

darkfader: int.from_byte and int.to_bytes should be rather quick.
bigints are stored as an array of uint32_t and all it has to do is copy some bytes. I chose big-endian was chosen because the data bits were in reverse as well (and this made masking/tweaking pattern order looked more logical as well).

I don't see a faster method for python2 right away. Perhaps suggest users to use python3 with a nice warning or automagically try python3...

```
if __name__ == "__main__":
    if sys.version_info[0] == 2:
        try:
            exit(os.execv('/usr/bin/env', ['env', 'python3', __file__] + sys.argv[1:]))
        except OSError:
            pass
    main()
```

darkfader: well.. big ints are there really, just not the conversion. So it might be easier just to implement these.
I found functions:

```
def to_bytes(n, length, endianess='big'):
    h = '%x' % n
    s = ('0'*(len(h) % 2) + h).zfill(length*2).decode('hex')
    return s if endianess == 'big' else s[::-1]

or a variant

def to_bytes(n, length, endianess='big'):
    h = '%x' % n
    s = codecs.decode(('0'*(len(h) % 2) + h).zfill(length*2), "hex_codec")
    return s if endianess == 'big' else s[::-1]
```
for int.to_bytes. Found some other code here https://gemfury.com/agriconnect/python:cryptography/-/content/utils.py

A slightly modified version here:
```
import binascii

if hasattr(int, "from_bytes"):
    int_from_bytes = int.from_bytes
else:
    def int_from_bytes(data, byteorder, signed=False):
        assert byteorder == 'big'
        assert not signed

        return int(binascii.hexlify(data), 16)


if hasattr(int, "to_bytes"):
    def int_to_bytes(integer, length=None, byteorder='big', signed=False):
        return integer.to_bytes(
            length or (integer.bit_length() + 7) // 8 or 1, byteorder, signed
        )
else:
    def int_to_bytes(integer, length=None, byteorder='big', signed=False):
        assert byteorder == 'big'
        assert not signed

        hex_string = '%x' % integer
        if length is None:
            n = len(hex_string)
        else:
            n = length * 2
        return binascii.unhexlify(hex_string.zfill(n + (n & 1)))
```
which seems to work.

tim-nordell-nimbelink: Note: You can also speed it up by taking advantage of large multiplications.  It's a minor speed improvement.

```
 @@ -295,6 +295,13 @@ def _flash_encryption_tweak_range_bits(flash_crypt_config=0xF):
     return tweak_range
 
 
+# Forward bit order masks
+mul1        = 0x0000200004000080000004000080001000000200004000080000040000800010
+mul2        = 0x0000000000000000200000000000000010000000000000002000000000000001
+
+mul1_mask   = 0xffffffffffffff801ffffffffffffff00ffffffffffffff81ffffffffffffff0
+mul2_mask   = 0x000000000000007fe00000000000000ff000000000000007e00000000000000f
+
 def _flash_encryption_tweak_key(key, offset, tweak_range):
     """Apply XOR "tweak" values to the key, derived from flash offset
     'offset'. This matches the ESP32 hardware flash encryption.
@@ -322,12 +329,9 @@ def _flash_encryption_tweak_key(key, offset, tweak_range):
 
     else:
         key = int.from_bytes(key, byteorder='big', signed=False)
-        for offset_bit in range(24):
-            if offset & (1 << offset_bit) != 0:
-                key ^= _FLASH_ENCRYPTION_TWEAK_PATTERN_REVERSE[offset_bit] & tweak_range
-
-        key = int.to_bytes(key, length=32, byteorder='big', signed=False)
-        return key
+        addr = offset >> 5
+        key ^= (mul1 * addr) | ((mul2 * addr) & mul2_mask)
+        return int.to_bytes(key, length=32, byteorder='big', signed=False)
 
 
 def generate_flash_encryption_key(args):
```

Additionally, the input_file.tell() call, at least in Linux, invokes a system call per invocation, but the file I/O is buffered in 4k chunks.  You can gain some more time by keeping track of the offset internally.  Take a look at the improvements to the speed I made: https://github.com/tim-nordell-nimbelink/esptool/tree/feature/encryption_speed

tim-nordell-nimbelink: https://github.com/darkfader/esptool/pull/1/files

I also suspect that it could be a little faster if we used gray code order for encoding, and did a differential xor from address to address generated in gray code order.  This would cause out of order accesses to the files, but might be an overall speed improvement, however, that'd be a bit more complicated.  You could also do chunks of the file in parallel to take advantage of multiple cores in a system, too.  (See https://en.wikipedia.org/wiki/Gray_code).

projectgus:
@darkfader @tim-nordell-nimbelink Sorry for delayed reply on this, I saw Travis was red and kept skimming over it.

Flake8 is just complaining due to a missing double-newline. I'll manually cherry-pick this now.

projectgus:
Squashed and cherry-picked as abfacc0, thanks again both of you.

PR #436, opened 2019-06-11
ESP Efuse : Get selected efuse value as JSON object

sbertrand-witekio:
# Description of change
Addition to get efuse value as JSON object for further processing

Linux, ESP32, WROOM, WROVER

# Comment

This allow to get efuse value so script can flash devices appropriately depending on secure boot or flash encryption state of the device.

### DIFF
--- a/espefuse.py
+++ b/espefuse.py
@@ -498,10 +498,17 @@ def dump(esp, _efuses, args):
 def summary(esp, efuses, args):
     """ Print a human-readable summary of efuse contents """
     ROW_FORMAT = "%-22s %-50s%s= %s %s %s"
-    print(ROW_FORMAT.replace("-50", "-12") % ("EFUSE_NAME", "Description", "", "[Meaningful Value]", "[Readable/Writeable]", "(Hex Value)"))
-    print("-" * 88)
+    json_efuse = {}
+    export_file=sys.stdout
+    if args.file:
+        print("Saving efuse values to " + args.file)
+        export_file = open(args.file, 'w')
+    if args.format == 'summary':
+        print(ROW_FORMAT.replace("-50", "-12") % ("EFUSE_NAME", "Description", "", "[Meaningful Value]", "[Readable/Writeable]", "(Hex Value)"),file = export_file)
+        print("-" * 88,file=export_file)
     for category in set(e.category for e in efuses):
-        print("%s fuses:" % category.title())
+        if args.format == 'summary':
+            print("%s fuses:" % category.title(),file=export_file)
         for e in (e for e in efuses if e.category == category):
             raw = e.get_raw()
             try:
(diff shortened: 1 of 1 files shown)


mahavirj:
CC @ghemal

projectgus:
@sbertrand-witekio Thanks for sending this PR. We've been discussing the need for something similar internally, specifically a machine-parseable output for the efuse values.

Rather than having a command that reads a single efuse as JSON, what about adding an option to the existing `summary` command like `--format=json` that dumps the summary of all efuses as a JSON dictionary rather than the current human-readable table?

The main advantage would be that a single JSON file can contain machine-parseable data about all efuses. This will be much more convenient for any use case that involves reading more than one efuse at a time, because espefuse.py doesn't have to re-connect to the bootloader on each invocation.

I look forward to hearing what you think about this idea.

sbertrand-witekio:
That would be fine. I was just concerned about processing time, efuses are read from the device one at a time or in blocks then parsed by Python ?
Can there be an option to export to a file as well ?

projectgus:
@sbertrand-witekio 

In terms of runtime, nearly all of the time is spent connecting to the ESP32. Fetching the efuse information from the chip is quite fast.

Are you interested in updating this PR along the lines of what was discussed?

Angus

sbertrand-witekio:
Hi,
I can update the PR with those changes
efuse summary :
- `[--format=json]` to get a json output
- `[--file=filename]` to get export in a file

I'll work on this this week.

sbertrand-witekio:
I have updated the PR.

espefuse.py
```diff
@@ -836,8 +863,10 @@ def add_force_write_always(p):
         help='Run espefuse.py {command} -h for additional help')
 
     subparsers.add_parser('dump', help='Dump raw hex values of all efuses')
-    subparsers.add_parser('summary',
+    p = subparsers.add_parser('summary',
                           help='Print human-readable summary of efuse values')
+    p.add_argument('--format', help='Name of efuse register get',choices=['summary','json'],default='summary')
+    p.add_argument('--file', help='File to save the json formated summary')
```
projectgus: You can have argparse handle the file management entirely by setting `type=argparse.FileType('w')` and `default=sys.stdout`, see https://docs.python.org/2/library/argparse.html#type - that way the result will be stdout if the argument is not passed in, and the file otherwise.

projectgus: Also, nitpick but help text doesn't match bheaviour: the summary will be saved here whether JSON formatted or not.

espefuse.py
```diff
@@ -836,8 +863,10 @@ def add_force_write_always(p):
         help='Run espefuse.py {command} -h for additional help')
 
     subparsers.add_parser('dump', help='Dump raw hex values of all efuses')
-    subparsers.add_parser('summary',
+    p = subparsers.add_parser('summary',
                           help='Print human-readable summary of efuse values')
+    p.add_argument('--format', help='Name of efuse register get',choices=['summary','json'],default='summary')
```
projectgus: Help text here is copy-paste.

espefuse.py
```diff
@@ -496,10 +498,17 @@ def dump(esp, _efuses, args):
...
     """ Print a human-readable summary of efuse contents """
     ROW_FORMAT = "%-22s %-50s%s= %s %s %s"
-    print(ROW_FORMAT.replace("-50", "-12") % ("EFUSE_NAME", "Description", "", "[Meaningful Value]", "[Readable/Writeable]", "(Hex Value)"))
-    print("-" * 88)
+    json_efuse = {}
+    export_file=sys.stdout
+    if args.file:
+        print("Saving efuse values to " + args.file)
+        export_file = open(args.file, 'w')
+    if args.format == 'summary':
```
projectgus: This is a nitpick and I don't mind if you don't change it, but it would be a little cleaner to set a local variable like `human_output = (args.format == 'summary')` and then `if human_output` in all these tests.

espefuse.py
```diff
@@ -515,26 +524,44 @@ def summary(esp, efuses, args):
...
+        if args.file:
+            export_file.close()
+            print("Done")
+    if args.format == 'json':
+        if args.file:
+            export_file.write(json.dumps(json_efuse))
+            export_file.close()
+            print("Done")
+        else:
+            print(json.dumps(json_efuse,sort_keys=True,indent=4))
```
projectgus: Suggest doing
```
json.dump(json_efuse, export_file, sort_keys=True, indent=4)
```
rather than doing something different for each case here.

It's less code, and formatting the file in a more friendly way may also be useful.

sbertrand-witekio:
I have fix the coding style based on the feedback.

projectgus:
Thanks @sbertrand-witekio !

PR #437, opened 2019-06-24
Fix too-broad-exception as discussed in PEP8

cclauss:
(Please delete any lines which don't apply)

# Description of change

### DIFF
--- a/pyaes/aes.py
+++ b/pyaes/aes.py
@@ -74,7 +74,7 @@ def _concat_list(a, b):
 # Python 3 compatibility
 try:
     xrange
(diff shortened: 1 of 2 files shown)


projectgus:
Hi @cclauss ,

As [mentioned on your other PR](https://github.com/espressif/esptool/pull/439#issuecomment-505282409), this library is vendored from an external source and we'll probably remove the vendored libraries at some point in the near(ish) future.

For this reason I don't really want to modify their code, even though it has some PEP8-non-compliant features in it.

Angus

Issue #339, opened 2018-07-31
Please include the test/ directory on PyPI


loader-bsd:
Please include the test/ directory in the release tarball on PyPi.
https://pypi.org/project/esptool/

projectgus:
Hi @loader-bsd ,

Thanks for getting in touch. I deliberately left this out because most users (rather than developers) of esptool won't need to use it, and it significantly increases the size of the package (because of all the binary images). May I ask why you need it in the package?

radimkarnis:
I’m closing this issue because it has been inactive for a few months and it seems like it's not relevant anymore. Please reopen if further discussion is needed.

PR #439, opened 2019-06-24
Actually run flake8 on the codebase

cclauss:
Perhaps run [__python/black__](https://github.com/python/black) on the codebase and then fix the other flake8 issues by hand.

(Please delete any lines which don't apply)

# Description of change

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,7 @@ script:
   - python test/test_imagegen.py
   - python test/test_espsecure.py
   - pip install flake8 flake8-future-import  # matches tests_require in setup.py, a bit hacky
+  - flake8 .
   - python setup.py flake8
   - python setup.py install
   - ( cd / && esptool.py --help )  # check the installed versions can run


projectgus:
Hi @cclauss,

Thanks for your interest in esptool & clean Python codebases.

Travis already runs flake8 on the source files which form part of the esptool package (note the `python setup.py flake8` line in the .travis.yml file, under the line added in this PR).

From the remaining code, most of the PEP8 failures are in the "vendored" libraries which are included in the repo to make it easier for users who don't want to use pip. Probably these libraries will be removed in the future anyhow, as I think more users are getting esptool via pip now compared to when these features were first introduced in 2016. In any case I don't think it's practical to modify vendored code to keep PEP8 happy.

The remaining code which is not covered by PEP8 are the automated tests (`test/` directory) and the short wrapper script which is used when rebuilding the flasher stub. A PR which made this code PEP8 compliant would be most welcome (and it would help with resolving #339). However it probably won't become a key priority for us to implement in the near future.

Angus

cclauss:
Makes sense.

PR #446, opened 2019-07-21
Minor python3 fix: bare print -> print()

leres:
Updating one of my projects I was happy to find python 3 is now supported.

There is a bare  print in esptool.py that needs to be converted to print()


projectgus:
Hi @leres ,

Thanks for that! I'm surprised Flake8 did not catch this, but I guess `function` is a valid no-op (references the function name).

PR #476, opened 2019-11-08
SPDX

yegorich:
# Description of change

This PR adds initial support for SPDX identifiers. This approach is intended to help making FOSS compliance process easier.


kaliGirl92277:
Why is merge not recognized as available function? Too many dependencies?

Issue #499, opened 2020-01-03
flasher_stub GCC 9 warnings/errors


paravoid:
Building flasher_stub (as of efb55ac15c14044c6e031357ea4ba5ae67dc8f39) with gcc version 9.2.1 20191130 (9.2.1-21+4) ([Debian's](https://packages.debian.org/sid/gcc-xtensa-lx106)), results into the following compiler warnings (that due to the stock Makefile's `-Werror` are converted into errors):

```
stub_write_flash.c:53:23: warning: 'STATUS_QIE_BIT' defined but not used [-Wunused-const-variable=]
   53 | static const uint32_t STATUS_QIE_BIT = (1 << 9);  /* Quad Enable */
      |                       ^~~~~~~~~~~~~~
miniz.c: In function 'tinfl_decompress':
miniz.c:1516:9: warning: this 'for' clause does not guard... [-Wmisleading-indentation]
 1516 |         for ( i = 0; i <= 143; ++i) *p++ = 8; for ( ; i <= 255; ++i) *p++ = 9; for ( ; i <= 279; ++i) *p++ = 7; for ( ; i <= 287; ++i) *p++ = 8;
      |         ^~~
miniz.c:1516:47: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'for'
 1516 |         for ( i = 0; i <= 143; ++i) *p++ = 8; for ( ; i <= 255; ++i) *p++ = 9; for ( ; i <= 279; ++i) *p++ = 7; for ( ; i <= 287; ++i) *p++ = 8;
      |                                               ^~~
miniz.c: In function 'tdefl_find_match':
miniz.c:2334:5: warning: this 'if' clause does not guard... [-Wmisleading-indentation]
 2334 |     if (!dist) break; p = s; q = d->m_dict + probe_pos; for (probe_len = 0; probe_len < max_match_len; probe_len++) if (*p++ != *q++) break;
      |     ^~
miniz.c:2334:23: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'if'
 2334 |     if (!dist) break; p = s; q = d->m_dict + probe_pos; for (probe_len = 0; probe_len < max_match_len; probe_len++) if (*p++ != *q++) break;
      |                       ^
  WRAP build/stub_flasher_8266.elf wrap_stub.py -> build/stub_flasher_snippet.py
```

[...]

PR #500, opened 2020-01-06
Update miniz to 2.1.0

paravoid:
Update miniz to the latest version, 2.1.0. This is the stock, amalgated
miniz, with a few modifications, forward-ported from the existing diff
against 1.5:
 * Configuration options (the definitions at the top of miniz.h)
 * Disabling assert(), i.e. defining MZ_ASSERT to nothing
 * Defining TDEFL_LESS_MEMORY to 1
 * Setting TDEFL_LZ_DICT_SIZE = 8*1024 and TDEFL_LZ_CODE_BUF_SIZE = 8 * 1024

An additional complication is that the new miniz performs a long long to
integer division, which requires code from libgcc (__ashldi3), which we
didn't link to by default (due to -nostdlib). Add -lgcc to LDLIBS.

This seems to be a small increase in .text (8018->8023) with the stock
toolchain, but with a newer one (gcc 9.2.1/picolibc 1.3) actually
results in a decrease (7990->7947).

Tested on real hardware (D1 Mini Pro).

See #499 for some additional context.


projectgus:
Hi @paravoid ,

Thanks for this, I hadn't actually noticed that miniz development was continuing on GitHub.

Angus

paravoid:
Thanks!

On that note, a two-parter idea to ease future migrations to newer versions:
- It'd be useful to expose the changes changes to separate config options: expose TDEFL_LESS_MEMORY as MINIZ_LESS_MEMORY, and maybe the DICT_SIZE/CODE_BUF_SIZE as MINIZ_EVEN_LESS_MEMORY or something? I don't feel confident enough to support those dict size changes and such, so perhaps it would make sense for you or someone this project to raise that with miniz upstream?
- I filed richgel999/miniz#145 to ask for the config to be split into a different header file. Hopefully that could allow us to commit an esptool-specific miniz_config.h separately in this project, and then just carry a pristine miniz.h/c without code changes in this project.

projectgus:
Cherry-picked as 18647f2, thanks @paravoid :)

projectgus:
@paravoid Just saw your suggestions. Agree these sound good, making miniz into a submodule or something would be a nice cleanup. Will see what I can do.

PR #501, opened 2020-01-06
Guard STATUS_QIE_BIT definition to ESP32/ESP32S2

paravoid:
Only used there, which causes a "defined but not used"
(-Wunused-const-variable) compiler error in newer versions of GCC.

Rather than disable what is otherwise a useful compiler warning, make
the definition conditional to the ESP32s.

See #499 for context.


projectgus:
Hi @paravoid ,

Thanks for this, change LGTM and good to know we can build the stub on gcc 9 with no other warnings.

Angus

projectgus:
Cherry-picked, thanks again @paravoid !

PR #518, opened 2020-03-27
Feature/serial port search espefuse (ESPTOOL-163)

lukeplatz:
# Description of change
A very useful feature of `esptool.py` is its ability to find an ESP Device even when no `--port` option is provided *AND* the connected Device is not on the default port.

This ability to scan ports until it finds a device was introduced in #307 and is [specifically noted as a feature in the esptool docs as a new behavior as of v2.4.0](https://github.com/espressif/esptool#common-options)

This feature is currently missing from `espefuse.py`. In order to keep the logic of connecting to ESP Devices consistent, I've created this PR which re-organizes the serial port scanning logic in `esptool.py` so that it is re-usable, and gone ahead and utilized it within `espefuse.py`

With this change, `espefuse.py` and `esptool.py` now behave the same when no port is provided as an argument.

- Operating Systems
  - macOS (10.15.3 - Catalina)
  - Windows 10
- Python 3.8.1
- Hardware
  - ESP32-D0WDQ6

```shell
./test_imagegen.py
............
Ran 12 tests in 1.828s

OK

./test_esptool.py /dev/cu.SLAB_USBtoUART esp32 230400 
Running esptool.py tests...
..s....sss..............................s.......
Ran 48 tests in 680.389s

OK (skipped=5)
```

### DIFF
--- a/espefuse.py
+++ b/espefuse.py
@@ -998,8 +998,8 @@ def main():
     operation_func = globals()[args.operation]
 
     # Connect to a device at given port, or iterate through ports if none provided
-    args.chip = 'auto' # set chip type to auto so connection searches for all devices
-    args.trace = False # set trace_enabled to false, this is the default for esptool.
+    args.chip = 'auto'  # set chip type to auto so connection searches for all devices
+    args.trace = False  # set trace_enabled to false, this is the default for esptool.
     esp = esptool.connect_to_esp(args, args.baud)
 
     # dict mapping register name to its efuse object
--- a/espefuse.py
+++ b/espefuse.py
@@ -903,7 +903,7 @@ def main():
     parser.add_argument(
         '--port', '-p',
         help='Serial port device',
-        default=os.environ.get('ESPTOOL_PORT', esptool.ESPLoader.DEFAULT_PORT))
+        default=os.environ.get('ESPTOOL_PORT', None))
 
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


projectgus:
Hi @lukeplatz ,

I'm really sorry I didn't get back to you when you submitted this in March.

I have some reservations about adding port detection to espefuse.py. A lot of the espefuse.py commands are destructive - they will perform one-way changes to the chip that cannot be reversed. For these commands I think it's important to know *exactly* what one is doing, and that extends to knowing which serial port you are interacting with and not finding out that there was a second board connected that you forgot about until just now.

There are some commands (like `summary` and `dump`) which are completely read-only and I can see these would be useful to support auto-detection. If you feel like updating the PR to support detection for a chosen set of commands only, then I think it could be a useful change. Unfortunately the espefuse.py code has changed a lot this year, and I'm guessing you may have also lost some interest over the past 7 months of being ignored (sorry again).

If you're still interested, or if you have any questions, then please let me know.

radimkarnis:
Hello @lukeplatz,

Thank you for your feature proposal.
As it was mentioned before, this feature might introduce a possibility to brick a board unintentionally in the case when you have multiple devices connected to COMs. 
I am therefore closing this PR, but you are more than welcome to submit a new one if the given feature would be just for read-only commands.

PR #546, opened 2020-07-21
Use segments for app image generation (ESPTOOL-126)

Skallwar:
# Description of change
This PR add a ```--use_segments``` to elf2image. This option use segments instead of sections in order to build the app image. This is usefull if there is more than 16 PROGBITS ELF sections (limit of app image segments).

[Here is a discusion about this](https://www.esp32.com/viewtopic.php?p=62918&sid=564af52d09a51a307f9e9ef9eba6eff6#p62918)

(Arch Linux (Linux 5.7.9), ESP32-DevKitC, ESP32-D0WDQ6)

```
....s......x...............................s.......
Ran 51 tests in 788.711s

OK (skipped=2, expected failures=1)
```

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -2361,7 +2361,6 @@ class ELFFile(object):
         self._read_sections(f, shoff, shnum, shstrndx)
         self._read_segments(f, _phoff, _phnum, shstrndx)
 
-
     def _read_sections(self, f, section_header_offs, section_header_count, shstrndx):
         f.seek(section_header_offs)
         len_bytes = section_header_count * self.LEN_SEC_HEADER
@@ -2398,7 +2397,6 @@ class ELFFile(object):
         def lookup_string(offs):
             raw = string_table[offs:]
             return raw[:raw.index(b'\x00')]
-
 
         prog_sections = [ELFSection(lookup_string(n_offs), lma, read_data(offs, size)) for (n_offs, _type, lma, size, offs) in prog_sections
                          if lma != 0 and size > 0]
@@ -2918,7 +2916,7 @@ def elf2image(args):
         image = ESP8266V2FirmwareImage()
     image.entrypoint = e.entrypoint
     if args.use_segments:
-        image.segments = e.segments # ELFSection is a subclass of ImageSegment
+        image.segments = e.segments  # ELFSection is a subclass of ImageSegment
     else:
         image.segments = e.sections  # ELFSection is a subclass of ImageSegment
     image.flash_mode = {'qio':0, 'qout':1, 'dio':2, 'dout': 3}[args.flash_mode]
--- a/esptool.py
+++ b/esptool.py
@@ -2324,6 +2324,9 @@ class ELFFile(object):
 
     LEN_SEC_HEADER = 0x28
 
+    SEG_TYPE_LOAD = 0x01
+    LEN_SEG_HEADER = 0x20
+
     def __init__(self, name):
         # Load sections from the ELF file
         self.name = name
@@ -2356,6 +2359,8 @@ class ELFFile(object):
         if shnum == 0:
             raise FatalError("%s has 0 section headers" % (self.name))
         self._read_sections(f, shoff, shnum, shstrndx)
+        self._read_segments(f, _phoff, _phnum, shstrndx)
+
 
     def _read_sections(self, f, section_header_offs, section_header_count, shstrndx):
         f.seek(section_header_offs)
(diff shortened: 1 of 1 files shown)


projectgus:
Hi @Skallwar ,

I just wanted to say sorry for the slow update on this. I haven't been totally ignoring you, instead have been trying to figure out if we can move to this for ESP-IDF & ESP32-S2 as well. The main sticking point on ESP32-S2 is the one I told you about, we need to start the IRAM segment at a dynamic address that depends on the length of the DRAM segment. I can't find a clean way to do this in the linker, it seems pretty determined that the start of the output section as defined in `MEMORY` should be the start of the segment in the ELF. Which is fair enough I guess, for most use cases.

After giving up on this I'm finishing up an option that automatically merges adjacent sections in the ELF instead, provided they land in the same memory type. This seems to work fine, I'm just updating the test cases in `test_imagegen.py` to match. Hoping to push a branch for you to take a look at, later this week.

Angus

projectgus:
Hi @Skallwar ,

Thanks again for being patient while I got time to look at this again.

I've pushed a Work In Progress commit that merges sections here: https://github.com/espressif/esptool/tree/feature/merge_adjacent_sections if you have a chance could you take a look and let me know if it works for your purposes?

There isn't anything wrong with the approach here, I agree technically it's more correct to use segments/phdrs for this purpose. But it looks like we may not have a way to make it work properly for ESP32-S2 and our other forthcoming chips, which means we'd always have to support both options. Merging sections is a way to avoid "option creep".

Angus

Skallwar:
> Thanks again for being patient while I got time to look at this again.

No problem.

> I've pushed a Work In Progress commit that merges sections here: https://github.com/espressif/esptool/tree/feature/merge_adjacent_sections if you have a chance could you take a look and let me know if it works for your purposes?

It's working for me. But I had to made some [changes](https://github.com/espressif/esptool/commit/a3c00b22f1d67fcc7bf2bdcddb6f377af07f43ea#r41851947)
 
> There isn't anything wrong with the approach here, I agree technically it's more correct to use segments/phdrs for this purpose. But it looks like we may not have a way to make it work properly for ESP32-S2 and our other forthcoming chips, which means we'd always have to support both options.

Does putting sections with *memory size* but no *physical size* in a separate segment could do the trick ?

> Merging sections is a way to avoid "option creep".

Yep it's a good compromise

projectgus:
Hi @Skallwar ,

Sorry this is crawling along so slowly, shortly after I replied to you I went on vacation.

> It's working for me. But I had to made some changes

My fault, sorry. A colleague suggested I use "all" here and I fat-fingered the change. Have pushed an amended commit on the branch.

We're currently have some internal discussion with our Zephyr devs about whether to merge your PR, the `feature/merge_adjacent_sections` branch, or if the best solution is this one: https://github.com/zephyrproject-rtos/zephyr/pull/27070 .

> Does putting sections with memory size but no physical size in a separate segment could do the trick ?

I've had another look and if I understand what you're suggesting correctly then maybe it can work, I hadn't tried combingin assignments to an output memory region and a phdr segment for each section.

Skallwar:
Hi Angus,

> Sorry this is crawling along so slowly, shortly after I replied to you I went on vacation.

No problem.

> We're currently have some internal discussion with our Zephyr devs about whether to merge your PR, the `feature/merge_adjacent_sections` branch, or if the best solution is this one: [zephyrproject-rtos/zephyr#27070](https://github.com/zephyrproject-rtos/zephyr/pull/27070).

In my opinion this is a temporary patch. If user code where to create more custom sections (or Zephyr itself), this will not hold very long.

> I've had another look and if I understand what you're suggesting correctly then maybe it can work, I hadn't tried combingin assignments to an output memory region and a phdr segment for each section.

Yes that exactly what I was suggesting, sorry to not have been clearer :sweat_smile:

Skallwar:
@projectgus Any news on this?

newblaque (approved):

projectgus:
@Skallwar sorry again for the long silence. Have squashed these commits and cherry-picked them into our internal review queue.

The one thing I would like to add is an image generation test case. Do you happen to have an ELF file and a corresponding (working) BIN file created with this option, that I can use for this purpose? (ELF file will end up committed in the repo, so ideally nothing sensitive!)

Skallwar:
This is for esp32 (zephyr hello_world sample):

[zephyr_phdr_test.tar.gz](https://github.com/espressif/esptool/files/5672517/zephyr_phdr_test.tar.gz)

projectgus:
@Skallwar Thanks for being so patient. This is cherry-picked in ca16d5f3ad471c4223e0895221325feb850b9582 with a few small tweaks for flake8, README note and tests added in 5f4deb052c6ea88d89a8836740a7d2547f718b1b and b96df73ba75cccd38ed6730829d8d01c0205e508

Skallwar:
I'm glad I could help a bit :)

> I've had another look and if I understand what you're suggesting correctly then maybe it can work, I hadn't tried combining assignments to an output memory region and a phdr segment for each section.

Out of curiosity, does this end up working for your use case?

projectgus:
> does this end up working for your use case?

@Skallwar To be honest, I haven't had time to look yet but after talking to our internal Zephyr team everyone agreed that it's a good option to have anyway (although I think our Zephyr folks found another way around it for now).

PR #552, opened 2020-08-26
create a fast modular exponent function

krishnajalan:
Faster method.
This function run in log(N) time complexity and log(N) space complexity to find the modular exponent and can be used cryptographic encryption to find the modular exponent of a really large prime number

### DIFF
--- a/ecdsa/numbertheory.py
+++ b/ecdsa/numbertheory.py
@@ -33,7 +33,7 @@ def modular_exp( base, exponent, modulus ):
   if exponent < 0:
     raise NegativeExponentError( "Negative exponents (%d) not allowed" \
                                  % exponent )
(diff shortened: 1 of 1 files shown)


projectgus:
Hi @krishnajalan ,

Nice work!

Have you considered submitting this to the [python ecdsa project instead](https://github.com/warner/python-ecdsa/)? Would prefer if it gets reviewed and merged there and then we can update the vendored library to use this version, rather than forking the upstream library.

krishnajalan:
thank for responding.
I will try submittting it to python ecdsa project.

projectgus:
It seems like the [upstream PR has been closed](https://github.com/warner/python-ecdsa/pull/209) so I'm going to close this as well.

PR #558, opened 2020-09-10
esptool_py: make python3 compatible

niziak:


projectgus:
Hi @niziak,

Thanks for sending this PR. Just to check: is the purpose of this change to avoid an exception if the same address is repeated twice in the argument list?

Issue #254, opened 2017-12-26
Command to merge multiple .bin files into one file for flashing (ESPTOOL-11)


earlt:
I am trying to create an image of the latest ESP8266_NONOS_SDK for an ESP8266-01 module using the latest esptool.py (v2.2) 

Flashing with this works fine:

esptool.py --baud 115200 --port /dev/ttyUSB0 write_flash -fm dio -ff 40m -fs detect \
0x00000 /opt/arduino/esp8266/ESP8266_NONOS_SDK-master/bin/boot_v1.7.bin \
0x01000 /opt/arduino/esp8266/ESP8266_NONOS_SDK-master/bin/at/512+512/user1.1024.new.2.bin \
0x81000 /opt/arduino/esp8266/ESP8266_NONOS_SDK-master/bin/at/512+512/user2.1024.new.2.bin \
0x7E000 /opt/arduino/esp8266/ESP8266_NONOS_SDK-master/bin/blank.bin \
0xFC000 /opt/arduino/esp8266/ESP8266_NONOS_SDK-master/bin/esp_init_data_default_v08.bin \
0xFE000 /opt/arduino/esp8266/ESP8266_NONOS_SDK-master/bin/blank.bin

However, when I create an image via:

esptool.py -c esp8266 make_image \
-f /opt/arduino/esp8266/ESP8266_NONOS_SDK-master/bin/boot_v1.7.bin -a 0x00000 \
-f /opt/arduino/esp8266/ESP8266_NONOS_SDK-master/bin/at/512+512/user1.1024.new.2.bin -a 0x01000 \
-f /opt/arduino/esp8266/ESP8266_NONOS_SDK-master/bin/at/512+512/user2.1024.new.2.bin -a 0x81000 \
-f /opt/arduino/esp8266/ESP8266_NONOS_SDK-master/bin/blank.bin -a 0x7E000 \
-f /opt/arduino/esp8266/ESP8266_NONOS_SDK-master/bin/esp_init_data_default_v08.bin -a 0xFC000 \
-f /opt/arduino/esp8266/ESP8266_NONOS_SDK-master/bin/blank.bin -a 0xFE000 \
ESP8266-Master

and try to flash the resultant image with:

esptool.py --baud 115200 --port /dev/ttyUSB0 write_flash -fm dio -ff 40m -fs detect 0x00 ESP8266-Master0x00000.bin

the ESP8266 does not respond. 

Is there something wrong with make_image?

projectgus:
[...]

PR #568, opened 2020-10-23
Adding merge_image functionality (ESPTOOL-79)

Barmaley13:
# Description of change
Added command to merge multiple .bin files into one file for flashing.

Furthermore, you can create both unencrypted and encrypted images (-k is optional)
`esptool.py --port PORT merge_image -k flash_key.bin @download.config`

and flash those with 0x0 offset, for instance
`esptool.py --port PORT -b 460800 write_flash --flash_mode dio --flash_freq 40m --flash_size detect 0x0 flash.bin`

or similar (for the encrypted flash)
`esptool.py --port PORT -b 460800 write_flash --flash_mode dio --flash_freq 40m --flash_size detect 0x0 flash.enc.bin`

#254 

Ubuntu 18.04
ESP32-WROOM-32D
ESP32

# I have NOT run the esptool.py automated integration tests with this change and the above hardware.
This addition would not affect automated integration tests since this pull request does not modify existing functionality it appends it.

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -3146,6 +3146,90 @@ def make_image(args):
     image.save(args.output)
 
 
+def merge_image(args):
+    # Helper classes
+    class Bin(object):
+        def __init__(self, addr, file):
+            self.addr = addr
+            self.file = file
+            self.file_name = os.path.basename(file.name)
+            self.size = self.get_size()
+
+        def get_size(self):
+            self.file.seek(0, 2)  # seek to end
+            size = self.file.tell()
+            self.file.seek(0)
+            return size
+
+    class MultiBin(object):
+        def __init__(self, name, output_folder='.'):
+            self.name = name
+            self.output_folder = output_folder
+            try:
+                os.makedirs(os.path.realpath(self.output_folder))
+            except:
+                pass
+            self.output_path = os.path.realpath(os.path.join(self.output_folder, self.name))
+            self.bin_array = []
+
+        def add_bin(self, addr, file):
+            self.bin_array.append(Bin(addr, file))
+
+        def sort_bin(self):
+            swapped = True
+            while swapped:
+                swapped = False
+                for i in range(len(self.bin_array) - 1):
+                    if self.bin_array[i].addr > self.bin_array[i + 1].addr:
+                        self.bin_array[i], self.bin_array[i + 1] = self.bin_array[i + 1], self.bin_array[i]
+                        swapped = True
+
+        def add_bin_to_other_bin(self, previous, binary):
+            with open(self.output_path, "a") as output_file:
+                output_file.write('\xff' * (binary.addr - previous))
+            print("Add %s from 0x%x to 0x%x (0x%x)" % (
+            binary.file_name, binary.addr, binary.addr + binary.size, binary.size))
+            with open(self.output_path, "a") as output_file:
+                output_file.write(binary.file.read())
+            return binary.addr + binary.size
+
+        def create_bin(self):
+            new_start = 0
+            open(self.output_path, "wb").close
+            for b in self.bin_array:
+                new_start = self.add_bin_to_other_bin(new_start, b)
+
+        def check_if_possible(self):
+            for i in range(1, len(self.bin_array)):
+                if (self.bin_array[i].addr <= (self.bin_array[i - 1].addr + self.bin_array[i - 1].size)):
+                    print(self.bin_array[i].addr, (self.bin_array[i - 1].addr + self.bin_array[i - 1].size))
+                    raise Exception("Not possible to create this bin, overlapping between %s and %s" % (
+                        self.bin_array[i].file_name, self.bin_array[i - 1].file_name))
+
+    # Check that bins are provided
+    if len(args.addr_filename) == 0:
+        raise FatalError('No segments specified')
+
+    encrypt = bool(args.keyfile is not None)
+    unenc_name = args.output + '.bin'
+    enc_name = args.output + '.enc.bin'
+
+    # Combine bins into unencrypted image
+    mb = MultiBin(unenc_name)
+    for address, argfile in args.addr_filename:
+        mb.add_bin(address, argfile)
+    mb.sort_bin()
+    mb.check_if_possible()
+    mb.create_bin()
+
+    # Encrypt image (if needed)
+    if encrypt:
+        with open(unenc_name, 'rb') as input_file:
+            with open(enc_name, 'wb') as output_file:
+                espsecure._flash_encryption_operation(
+                    output_file, input_file, 0, args.keyfile, args.flash_crypt_conf, False)
+
+
 def elf2image(args):
     e = ELFFile(args.input)
     if args.chip == 'auto':  # Default to ESP8266 for backwards compatibility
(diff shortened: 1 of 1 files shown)


esptool.py
```diff
@@ -3075,6 +3115,92 @@ def make_image(args):
...
+        mb.add_bin(address, argfile)
+    mb.sort_bin()
+    mb.check_if_possible()
+    mb.create_bin()
+
+    # Encrypt image (if needed)
+    if encrypt:
+        with open(unenc_name, 'rb') as input_file:
+            with open(enc_name, 'wb') as output_file:
+                espsecure._flash_encryption_operation(
```
projectgus, line 3200: the underscore prefix in python is for private methods. overall, as this command will probably be run as part of a script or a custom build process then I'd prefer to make encryption a separate step in the script - otherwise when we add ESP32-S2 flash encryption support (for example) this command will also need updating to support that.

esptool.py
```diff
@@ -3075,6 +3115,92 @@ def make_image(args):
...
+                os.makedirs(os.path.realpath(self.output_folder))
+            except OSError:
+                pass    # file already exists, overwrite
+            self.output_path = os.path.realpath(os.path.join(self.output_folder, self.name))
+            self.bin_array = []
+
+        def add_bin(self, addr, file):
+            self.bin_array.append(Bin(addr, file))
+
+        def sort_bin(self):
```
projectgus, line 3149: bubble sort! blast from the past. :)

Python can you help you here, though: `self.bin_array = sorted(self.bin_array, key=lambda b:b.addr)`

esptool.py
```diff
@@ -3075,6 +3115,92 @@ def make_image(args):
...
+    class MultiBin(object):
+        def __init__(self, name, output_folder='.'):
+            self.name = name
+            self.output_folder = output_folder
+            try:
+                os.makedirs(os.path.realpath(self.output_folder))
+            except OSError:
+                pass    # file already exists, overwrite
+            self.output_path = os.path.realpath(os.path.join(self.output_folder, self.name))
+            self.bin_array = []
```
projectgus, line 3144: this is a bit of a nitpick, but Python doesn't have arrays - this a list. Suggest calling it `bins` not `bin_array`.

esptool.py
```diff
@@ -3075,6 +3115,92 @@ def make_image(args):
...
+        def create_bin(self):
+            new_start = 0
+            open(self.output_path, "wb").close()
+            for b in self.bin_array:
+                new_start = self.add_bin_to_other_bin(new_start, b)
+
+        def check_if_possible(self):
+            for i in range(1, len(self.bin_array)):
+                if self.bin_array[i].addr <= (self.bin_array[i - 1].addr + self.bin_array[i - 1].size):
+                    print(self.bin_array[i].addr, (self.bin_array[i - 1].addr + self.bin_array[i - 1].size))
```
projectgus, line 3176: debug statement left in by mistake?

It should be possible to use the `AddrFilenamePairAction` which is already used by `write_flash` command to test none of the binaries overlap.

esptool.py
```diff
@@ -3075,6 +3115,92 @@ def make_image(args):
...
+                output_file.write('\xff' * (binary.addr - previous))
+            print("Add %s from 0x%x to 0x%x (0x%x)" % (
+                binary.file_name, binary.addr, binary.addr + binary.size, binary.size))
+            with open(self.output_path, "a") as output_file:
+                output_file.write(binary.file.read())
+            return binary.addr + binary.size
+
+        def create_bin(self):
+            new_start = 0
+            open(self.output_path, "wb").close()
```
projectgus, line 3169: why open a file, immediately close it, then open it append to it, close it multiple times - when you could just open it once and write into it multiple times?

projectgus (changes requested):
Hi @Barmaley13,

I'm sorry noone has gotten back to you for such a long time.

Thanks for contributing this, it's a useful function and something that's been obviously missing from esptool.py for some time.

This PR needs some work before we can accept it. If you're interested in reworking it then I can promise a < 2 month turnaround from now on - although if you'd rather not then that's OK as well.

As well as the inline comments, some other things:

* The PR is showing a lot of unrelated changes (I just reviewed the `merge_image` function not the other code as it was mostly merge-related issues). Please rebase on the master branch to clean this up.
* The PR is failing some of the automated Python checks (flake8, etc). Please check those out.
* It would be great to have some unit tests that the output is correct, errors correctly detected, etc.
* The `--flash_size`/`--flash_mode` options are actually not applied unless you flash a binary directly at the bootloader offset (so 0x1000 for ESP32). So this part won't quite work as described in the PR. It'd be great if these options could be supplied to this command as well, in the same way they can be supplied d to `write_flash` command, so that the output file is generated with the specified modes/etc.

esptool.py
Barmaley13: Renamed!

esptool.py
Barmaley13: Updated per your suggestion

esptool.py
Barmaley13: I've cleaned up code so it does it only once. That was not my code to begin with but hopefully this latest version makes more sense.

esptool.py
Barmaley13: Removed debug statement. I've also cleaned up overlap check. But by no means if we have this functionality else where we should use it rather than rewrite it.

esptool.py
Barmaley13: You are probably right! I am going to leave it as it is for now. But maybe you can change it to serve esptool needs better

Barmaley13:
Hi @projectgus,
Thank you for looking at this!
Yes, I took someone else's code that was not part of the esptool and glued things together to make it work since I needed that
functionality myself. I think I'd rather have someone else to finish this effort since I know so much about esptool and still somewhat amateur at contributing to open source projects.

> As well as the inline comments, some other things:
> 
> * The PR is showing a lot of unrelated changes (I just reviewed the `merge_image` function not the other code as it was mostly merge-related issues). Please rebase on the master branch to clean this up.

Sorry, I have not used rebase much. I think I just did it though

> * The PR is failing some of the automated Python checks (flake8, etc). Please check those out.

Just finished cleaning up portion of the code. Will see what kind of suggestions travis will come up with

> * It would be great to have some unit tests that the output is correct, errors correctly detected, etc.

I would love to leave this for someone else. I am hoping this latest version is good enough for handing it over to someone else to polish it up to the finish line.

> * The `--flash_size`/`--flash_mode` options are actually not applied unless you flash a binary directly at the bootloader offset (so 0x1000 for ESP32). So this part won't quite work as described in the PR. It'd be great if these options could be supplied to this command as well, in the same way they can be supplied d to `write_flash` command, so that the output file is generated with the specified modes/etc.

Going to leave this as is. Since I am not sure if this is done as a binary file offset or dynamically at flash time

positron96:
Hi guys. Is anywork being done for this feature? This will be very useful in CI/CD pipelines that automatically build firmwares

projectgus:
Hi @positron96,

I'm working on this but I need to write some tests for it, hopefully it'll be in the master branch in the next week or two. This PR thread will be updated once that happens.

Angus

PR #623, opened 2021-04-23
Fix to properly extract SecureBoot v2 public key (ESPTOOL-231)

gnought:
Fix the type error: a bytes-like object is required, not '_io.BufferedReader'


dobairoland:
Thank you for your contribution. We will look into this and contact you if there are any questions.

PR #625, opened 2021-04-25
Add build script (ESPTOOL-233)

kristofmulier:
<h1>Description of change</h1>

I added a build script in the top-level folder: `build_script.py`. This build script compiles the following Python scripts into executables:
- `espefuse.py` &nbsp;&nbsp;-> `build/espefuse(.exe)`
- `espsecure.py` -> `build/espsecure(.exe)`
- `esptool.py` &nbsp;&nbsp;&nbsp;&nbsp;-> `build/esptool(.exe)`

The build script uses **cx_Freeze** (https://cx-freeze.readthedocs.io) for this operation. It produces a `'build/'` subfolder containing an executable for each of the compiled scripts, along with the shared libraries (DLLs or .so files) needed to run them. The target computer doesn't need Python anymore.

For more information about the build script and optional arguments, just run it with the `-h` or `--help` flag.

<h1>Reason for change</h1>

Not all target computers (especially Windows) have Python pre-installed. Precompiled executables can therefore be interesting for such target computers.

Even target computers that *do* have a Python installation, can suffer from incompatibilities (eg. the dreaded `serial` vs `pyserial` issue). For these computers, precompiled executables can save a lot of hassle with installing and activating virtual environments.

<h1>Operating systems and Python versions</h1>

This build script works on both Windows and Linux. Only on Windows, the executables get the `'.exe'` suffix. The build script was tested on `Python 3.8`, but should work on anything above `Python 3.6`. Note that I'm using f-strings, which were introduced in `Python 3.6`.

The script will **not** work on `Python 2.x`.

<h1>Notes</h1>

Please feel free to make changes in my code. Feel free as well to contact me: `kristof<at>embeetle<dot>com`


radimkarnis:
Hello @kristofmulier,
thank you for your contribution and a thorough description. 
There is an ongoing internal discussion about this matter, we will get back to you as soon as possible.

kristofmulier:
Thank you @radimkarnis .

radimkarnis:
Hello @kristofmulier,
thanks again for the PR and your interest in esptool, your contribution is valued. 

We fully support the idea of building self-contained binaries. Nevertheless, the consensus is we would rather use `PyOxidizer` for Python packaging. `cx_freeze` produces binaries that have a similar signature as viruses and get often blocked by AV. That is a frequent problem in `arduino-esp32` right now, any time the esptool.exe (bundled by `PyInstaller`) is updated then someone has to get the binary whitelisted with various AV tools.

The plan is also to move the binary generation into the esptool project's CI. Then we can sign them and distribute with esptool releases.

If you're feeling super generous and interested in submitting another PR, we'd like to merge that; otherwise, we plan to work on this ourselves.
Radim

kristofmulier:
Hi @radimkarnis ,
Thank you very much for your reply. My apologies for this late reaction.

I didn't know that `cx_freeze` and `PyInstaller` binaries have this false-positive problem more often than `PyOxidizer`. Could it be because `PyOxidizer` uses Rust in the backend? I think most viruses are created in C/C++, so it makes sense that Rust-based applications are less likely to be flagged.

### Our experience with false positives

I'm part of the startup [Embeetle IDE](https://embeetle.com) - we develop a free IDE for microcontrollers. Our application is not based on any existing IDE (Eclipse, VSCode, ...) but really built up from the ground. We used Python for the GUI and C++ for the backend. To package the Python code, we use `cx_freeze`.

Unfortunately, our application regularly had false-positive flags on Windows. Even worse, Norton Antivirus would simply start deleting executables and `.pyd` files from our application right after a download! That's why we decided to buy an [EV Code Signing Certificate from Sectigo](https://sectigo.com/ssl-certificates-tls/code-signing) - "EV" stands for "Extended Verification". The company called us a few times to ask questions, reviewed our company's documents and finally approved our case. They sent us an USB-thumbdrive with a key. We use that to code-sign all executables and `.pyd` files in our application before each release, which I've automated as described in [this StackOverflow answer](https://stackoverflow.com/questions/31637875/advantages-with-code-signing-ev-certificates-other-than-the-smartscreen-ones/65630107#65630107).

The Code Signing certainly helps. Norton Antivirus and Windows might give a few warnings still, but they no longer block or delete our files.

Nevertheless, I'm still worried. We simply don't have the capacity or manpower to verify all other Antivirus software around. We'll only know if some other Antivirus - eg. McAfee, BitDefender, ... blocks our application if a user informs us, usually when it's alredy too late and we lost many potential users.

At some point we contemplated on buying a [VirusTotal](https://www.virustotal.com/gui/) license. They have this automated service that checks your binaries against tens or hundreds of common Antivirus softwares, to inform you if there are any false positives. Unfortunately, that service is crazy expensive (in the order of 500€ / month). We're a small team and don't have that kind of money.

How do you at Espressif deal with all of this? Do you rely on a service like [VirusTotal](https://www.virustotal.com/gui/)? Do you code-sign your applications? Or is the move towards `PyOxidizer` building enough to overcome all of this?

Thank you so much for your feedback.
Kind regards,
Kristof

*PS: We're in the midst of an effort to support the ESP chip family in our IDE. If you agree, I'll add you on LinkedIn and keep you updated.*

kristofmulier:
@radimkarnis ,
I gave it some thought, and had a look into my agenda. I will make time today and tomorrow to look into `PyOxidizer` with the aim to make a build script for `esptool`. I'll keep you updated about the progress.

PR #634, opened 2021-06-20
start with python3 as default (ESPTOOL-257)

rundekugel:
# Description of change
1st line let's start with python3 by default

"Pyserial is not installed for /usr/bin/python" 

ESP8285 & Ubuntu20

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 #
 # ESP8266 & ESP32 ROM Bootloader Utility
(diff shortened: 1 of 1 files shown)


dobairoland:
Hi @rundekugel. Thank you for the contribution but I'm sorry and we cannot accept it. Esptool still supports Python 2.7 even it reached its official end-of-life. Still a lot of users are stuck with older Pythons, e.g. old machines used only for flashing and nothing else.

We will make some coordinated drop of old versions in the near future. 

Thank you in advance for your understanding.

PR #650, opened 2021-08-09
fix comparison when handling deflated data (ESPTOOL-287)

alvarolb:
handle_flash_data should not be called if status < TINFL_STATUS_DONE, which meas decompress failed


radimkarnis:
Hi @alvarolb,

Thanks for your contribution, LTGM. No one has noticed this error for over 5 years!

Radim

alvarolb:
I suppose no one noticed the error as it is a little bug that in the wort case will do one unnecessary write on the flash... but nothing else ;)

PR #688, opened 2021-11-03
removed a default mutable argument pitfall (ESPTOOL-344)

NaelsonDouglas:
**The problem**
In Python it usually dangerous to use mutable arguments like dicts or lists as default arguments in methods, as is better explained [here](https://docs.python-guide.org/writing/gotchas/)

**the solution**
This PR applies a simple refactoring to remove a case where a method was created using default a mutable argument

### DIFF
--- a/esptool.py
+++ b/esptool.py
@@ -1865,7 +1865,9 @@ class ESP32S2ROM(ESP32ROM):
         return any(p == self.PURPOSE_VAL_XTS_AES256_KEY_1 for p in purposes) \
             and any(p == self.PURPOSE_VAL_XTS_AES256_KEY_2 for p in purposes)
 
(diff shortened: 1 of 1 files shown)


radimkarnis:
Hello @NaelsonDouglas, 
thank you for opening this PR, every contribution is highly valued!

In this case, this behavior is intended and used as a simple caching mechanism. The article you’ve provided explains this really well:

> When the Gotcha Isn’t a Gotcha:
Sometimes you can specifically “exploit” (read: use as intended) this behavior to maintain state between calls of a function. This is often done when writing a caching function.

I am sorry, but your changes would break this caching function and for this reason this PR can’t be accepted.

You are welcome to contribute in any other way!

Radim

PR #742, opened 2022-05-17
current script is located at __file__, not sys.executable (ESPTOOL-456)

dotlambda:
# Description of change
In order to remove the current script directory from `sys.path`, we have to remove `os.path.dirname(__file__)`. `sys.executable` is the location of the Python interpreter, not the script.

NixOS, no hardware needed

```
.............
Ran 13 tests in 2.788s

OK
Running espsecure tests...
Using espsecure 4.0 at /nix/store/gj41j6vrf4w7cwmsg064c8qigp5xfz2c-esptool-4.0/lib/python3.9/site-packages/espsecure/__init__.py
.....................
Ran 21 tests in 2.036s

OK
........
Ran 8 tests in 1.755s

OK
Running python modules tests...
.
Ran 1 test in 0.003s

OK
```

### DIFF
--- a/espefuse.py
+++ b/espefuse.py
@@ -20,7 +20,7 @@ import sys
 
 with contextlib.suppress(ValueError):
     if os.name != "nt":
-        sys.path.remove(os.path.dirname(sys.executable))
+        sys.path.remove(os.path.realpath(os.path.dirname(__file__)))
 
 import espefuse
 
--- a/espsecure.py
+++ b/espsecure.py
@@ -20,7 +20,7 @@ import sys
 
 with contextlib.suppress(ValueError):
     if os.name != "nt":
-        sys.path.remove(os.path.dirname(sys.executable))
+        sys.path.remove(os.path.realpath(os.path.dirname(__file__)))
 
 import espsecure
 
--- a/esptool.py
+++ b/esptool.py
@@ -19,7 +19,7 @@ import sys
 
 with contextlib.suppress(ValueError):
     if os.name != "nt":
-        sys.path.remove(os.path.dirname(sys.executable))
+        sys.path.remove(os.path.realpath(os.path.dirname(__file__)))
 
 import esptool
 


radimkarnis:
Hello @dotlambda,
thanks a lot for contributing.

Does this change solve any issues in your environment? This seems to break `esptool` on MacOS and Ubuntu linux (see [this log](https://github.com/espressif/esptool/runs/6484286825?check_suite_focus=true)).

The script directory isn't the issue here. What doesn't work is when the `esptool.py` wrapper script also gets installed into the **bin directory** of the virtual environment. This happens when `scripts=['esptool.py']` is specified in setup.py.

If I understand correctly the following happens in this case:

- you run esptool.py
- <virtualenv>/bin/esptool.py wrapper script gets executed
- it loads and execs <path to esptool package>/esptool.py
- ... which tries to import esptool
- which imports esptool.py wrapper script from <virtualenv>/bin/esptool.py

dotlambda:
> Does this change solve any issues in your environment?

Yes, this does solve my issue. The current solution only coincidentally works because `sys.executable` and `__file__` are both in `<virtualenv>/bin`, which is not the case in my environment.



> This seems to break `esptool` on MacOS and Ubuntu linux (see [this log](https://github.com/espressif/esptool/runs/6484286825?check_suite_focus=true)).

I can't reproduce that in a virtualenv, can you? This seems to boil down to the GitHub action calling `exec(compile(f.read(), __file__, 'exec'))` instead of properly starting a new Python process.

igrr:
> I can't reproduce that in a virtualenv, can you? This seems to boil down to the GitHub action calling exec(compile(f.read(), __file__, 'exec')) instead of properly starting a new Python process.

I can reproduce this. This isn't a Github action thing, I think, it is what `pip` does when you try to install the package in editable mode.

When we do `pip install esptool`, we get the contents of the actual esptool.py script from the root of this repository in `${VIRTUAL_ENV}/bin/esptool.py}`:
<details><summary>$ cat ${VIRTUAL_ENV}/bin/esptool.py</summary>

```py
#!/private/tmp/venv/bin/python3.9
#
# SPDX-FileCopyrightText: 2014-2022 Fredrik Ahlberg, Angus Gratton,
# Espressif Systems (Shanghai) CO LTD, other contributors as noted.
#
# SPDX-License-Identifier: GPL-2.0-or-later

# This executable script is a thin wrapper around the main functionality
# in the esptool Python package
#
# If esptool is installed via setup.py or pip then this file is not used at all,
# it's compatibility for the older "run from source dir" esptool approach.

# Linux/macOS: remove current script directory to avoid importing this file
# as a module; we want to import the installed esptool module instead
import contextlib
import os
import sys

with contextlib.suppress(ValueError):
    if os.name != "nt":
        sys.path.remove(os.path.dirname(sys.executable))

import esptool

if __name__ == "__main__":
    esptool._main()
```
</details>

However when installing a package in editable mode (`pip install -e esptool`), pip creates a wrapper script which loads the real script content from the package directory:
<details><summary>$ cat ${VIRTUAL_ENV}/bin/esptool.py</summary>

```py
#!/private/tmp/venv/bin/python3.9
# EASY-INSTALL-DEV-SCRIPT: 'esptool==4.1.dev0','esptool.py'
__requires__ = 'esptool==4.1.dev0'
__import__('pkg_resources').require('esptool==4.1.dev0')
__file__ = '/Users/ivan/e/esptool-py/esptool.py'
with open(__file__) as f:
    exec(compile(f.read(), __file__, 'exec'))
```
</details>

In this (editable) case, when esptool.py is invoked, `__file__` is the location e.g. in the checked out Git repository of esptool. However what we need to exclude from the search path is the location of the wrapper script created by `pip`.

On Linux and macOS systems we have seen, this has always been `${VIRTUAL_ENV}/bin/esptool.py`. It seems that the scripts are installed in a different location on your system (not under `bin`). Can you please check what this location is? (I have tried to find the answer to this question on NixOS online, but I couldn't find anything conclusive, sorry.)

dotlambda:
> It seems that the scripts are installed in a different location on your system (not under `bin`). Can you please check what this location is?

It's e.g. `/nix/store/gj41j6vrf4w7cwmsg064c8qigp5xfz2c-esptool-4.0/bin`, the hash changes.

But NixOS shouldn't be the only system where Python interpreter and `esptoool.py` aren't in the same directory. IIRC `pip install --user esptool` puts the latter in `~/.local/bin` whereas the interpreter would still be at e.g. `/usr/bin/python3`.



> This isn't a Github action thing, I think, it is what `pip` does when you try to install the package in editable mode.

How about making the GitHub action not use editable mode?

dotlambda:
> However what we need to exclude from the search path is the location of the wrapper script created by `pip`.

I actually think `pip` should be doing that by itself: it should replace the directory of the wrapper with the directory of the actual script.
Maybe it makes sense to file a PR upstream.

igrr:
Thank you for the extra details and for the reminder about the --user install! Agree both should be fixed, and we will add CI checks for different types of installation to prevent future regressions.

That said, editable install is also one of the use cases which we aim to support, so simply changing the Github action won't be an acceptable option.

I'll take a look if there is any reliable way to address all these use cases and will get back to you!

PR #770, opened 2022-08-23
C3 has BLE as feature (ESPTOOL-501)

Jason2866:
Info from get_chip_features is missing "BLE" for the C3


radimkarnis:
Hello @Jason2866,

thanks for the fix! Somehow, this got forgotten.

Issue #779, opened 2022-09-20
Faster CPU clock speed in stub flasher for higher performance (ESPTOOL-519)


Emill:
### Is your feature request related to a problem?

I'm mainly programming for ESP32-S3 with the built-in serial usb functionality.

The esptool.py flash performance is quite slow compared to the ideal when considering the USB speed and the erase/programming time for the flash.

A project I built flashes in 746 kbit/s.

### Describe the solution you'd like

I then modified the stub flasher initialisation in the following way:

1. Set SYSTEM_CPUPERIOD_SEL to 2.
2. Set SYSTEM_SOC_CLK_SEL to 1.

This sets the CPU frequency to 240 MHz instead of the default 20 MHz.

Now the project flashes at an average speed of 1405 kbit/s.

Hash verify time is also much faster.

Total time goes from 34 seconds to 16 seconds.

Have you considered increasing the cpu frequency in esptool, to speed up the flashing performance?

radimkarnis:
Hi @Emill,

Thanks for bringing this up, that's an interesting idea! 

You explain you have a working implementation of this, would you be willing to send a pull request? We can explore this idea more.

PR #781, opened 2022-09-20
Flasher stub: Increase CPU frequency to 240 MHz for ESP32-S3 (ESPTOOL-521)

Emill:
# Description of change

Increase clock speeds for ESP32-S3.

This increases CPU frequency from 20 MHz to 240 MHz and SPI frequency from 5 MHz to 20 MHz.

This results in ~2x faster flashing performance in total.

The power consumption impact has not been studied.

(https://github.com/espressif/esptool/issues/779), but only for ESP32-S3. Someone who has access to other boards can probably submit a similar patch for such boards.

ESP32-S3-WROOM-1

```
Running esptool.py tests...
...s....sss..........................s.....s.s...s.......s......
Ran 64 tests in 526.718s

OK (skipped=9)
```


igrr:
One note is that we should restore the original setting when exiting the stub (e.g. if `load_ram` is used). Otherwise the apps built before this feature is introduced will unexpectedly start running at a higher CPU and APB frequency.

radimkarnis:
@Emill, 

could you please update the PR to first store the original settings before increasing CPU freq and to restore the original settings when exiting the stub? (commands in question: `ESP_FLASH_END`, `ESP_FLASH_DEFLATED_END`, `ESP_MEM_END`, `ESP_RUN_USER_CODE`)

If not, I can finish this feature, please let me know!

Emill:
Do you think there is any issues regarding power consumption (peak power for example) of the increased cpu frequency?

radimkarnis:
The only issue I can think of might be the chip becoming more prone to browning out if an insufficient power supply is used.

Emill:
I was more thinking of people who try to flash a board that runs on a weak power supply, like a battery or something.

radimkarnis:
@Emill yes, I believe we are mentioning the same thing. I think we can implement this and see if any serious issues with weaker power supplies arise.

Emill:
I can revert the previous cpu setting whenever `stub_main` returns.

When `ESP_FLASH_END` etc. ends in a `software_reset()`, do I need to revert the settings there as well or will a reset automatically reset the cpu frequency?

For, `ESP_MEM_END`, should the cpu settings be reverted while `entrypoint_fn` is running?

radimkarnis:
We need to restore the settings in all of these four cases, the CPU freq can be increased only while stub logic is running.

radimkarnis:
Hi @Emill,
are you planning to work on this or should I pick this up? 😊 thank you and please let me know!

PR #802, opened 2022-11-28
[backport] fix(espefuse): Fix compatibility with Bitstring>=4 (ESPTOOL-552)

tjni:
This is a request to backport https://github.com/espressif/esptool/commit/ee27a6437576797d5f58c31e1c39f3a232a71df0 to the v3 branch.

I am updating `bitstring` in [nixpkgs](https://github.com/NixOS/nixpkgs) to v4.0.1 and would like to depend on it everywhere. nixpkgs packages both esptool v3 and v4 because (it seems) `esphome` still depends on v3.

Thank you!

### DIFF
--- a/espressif/efuse/emulate_efuse_controller_base.py
+++ b/espressif/efuse/emulate_efuse_controller_base.py
@@ -9,7 +9,13 @@ from __future__ import division, print_function
 
 import re
 
-from bitstring import BitString
+from bitstring import BitStream
+
+try:
+    FileNotFoundError
+except NameError:
+    # Python 2.7 compatibility
+    FileNotFoundError = IOError
 
 
 class EmulateEfuseControllerBase(object):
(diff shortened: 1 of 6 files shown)


dobairoland:
Hi @tjni. Thank you for the contribution. 

We have noticed some issues with our checks and fixed them in the `release/v3` branch. Could you please rebase on top of it to check your PR?

The main issue here could be that esptool in the master branch supports Python 3.7+ but in the `release/v3` we still keep compatibility with 2.7 and 3.4+. The latest version of `bitstring`  supports also just 3.7+. So we have to ensure that `bistring>=4` is not installed for incompatible Python version. 

The code also has to be compatible with 2.7 and 3.4+.

espressif/efuse/emulate_efuse_controller_base.py
```diff
@@ -28,15 +28,18 @@ def __init__(self, efuse_file=None, debug=False):
         self.efuse_file = efuse_file
         if self.efuse_file:
             try:
-                self.mem = BitString(open(self.efuse_file, 'a+b'), length=self.REGS.EFUSE_MEM_SIZE * 8)
-            except ValueError:
+                self.mem = BitStream(
+                    bytes = open(self.efuse_file, "rb").read(),
+                    length=self.REGS.EFUSE_MEM_SIZE * 8,
+                )
+            except (ValueError, FileNotFoundError):
```
dobairoland, line 35: For example, `FileNotFoundError` is not known on Python 2.7. I expect the PR check will fail on this line.

setup.py
```diff
@@ -120,7 +120,7 @@ def find_version(*file_paths):
         ],
     },
     install_requires=[
-        'bitstring>=3.1.6,<4',
+        'bitstring>=3.1.6',
```
dobairoland, line 123: Based on my above comment, maybe the following would be the cleanest and self-explaining way to avoid issues with some old `pip` versions:
```python
'bitstring>=3.1.6' if sys.version_info[:2] >= (3, 7) else 'bitstring>=3.1.6,<4',
```

(up for discussion)

espressif/efuse/emulate_efuse_controller_base.py
tjni: Thank you for bringing this up. I realize I need to do more thinking about this, so I will and come back with some ideas.

dobairoland: Probably something similar could help:
```python
try:
    FileNotFoundError
except NameError:
    FileNotFoundError = IOError
```

tjni: This is a great solution. I'm going to pull it in and see what CI says.

setup.py
tjni: I'm giving environment markers a try:

```
'bitstring>=3.1.6; python_version>="3.7"',
'bitstring>=3.1.6,<4; python_version<"3.7"',
```

Upon reflection, I do agree that it's important to keep using the old bitstream version on unsupported Python versions, as it can be a big headache to maintain otherwise.

radimkarnis:
Hi @tjni,
thanks for finishing the PR, the CI now passes. I've put this through our internal review queue and will merge if all goes well!

dobairoland:
Thank you @tjni for the contribution.

PR #822, opened 2023-01-11
refactor: Add python3 header to avoid python version confusion (ESPTOOL-589)

Manah7:
./esptool.py fail on my python installation due to python version confusing. Add `/usr/bin/env python3` avoid this error.

# Description of change
Refactor done with bash one-liner :
```bash
grep -rl '/usr/bin/env python' . | xargs sed -i s@'/usr/bin/env python'@'/usr/bin/env python3'@g
```

### DIFF
--- a/esp_rfc2217_server.py
+++ b/esp_rfc2217_server.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # SPDX-FileCopyrightText: 2009-2015 Chris Liechti
 # SPDX-FileContributor: 2020-2022 Espressif Systems (Shanghai) CO LTD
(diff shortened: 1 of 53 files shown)


dobairoland:
Hi @Manah7. Thank you for the contribution but I'm afraid that this will break other systems.

For example, in ESP-IDF environment (and other dependent frameworks and IDEs) `python` is the right one to include. There were the dependencies installed. On Windows it will right away error out because the environment installed for ESP-IDF doesn't even have a `python3`.

I would even suggest to run it as `python3 -m esptool` as `python3 ./esptool.py` which avoids any confusion.

Also many system offer ways to have python as python3. E.g. ubuntu/debian have a package called `python-is-python3`.

dobairoland:
No need to close this. I think some kind of fix we should incorporate.

These aliases might help in cases when there is no `python3`: https://github.com/espressif/esp-idf/blob/master/export.bat#L50

PR #841, opened 2023-02-16
espefuse: new argument for passing encryption key in command line instead of file path (ESPTOOL-621)

ysba:
# Description of change
Added a new command parameter in espefuse for passing the encryption key as argument instead of using a file on disk. The parametere is named ``--text-key``. The purpose of this change is to provide firmware already encrypted to the production environment. The encryption key is passed as argument to esptool by a C++ software. To burn the fuses and flash the firmware, a single button needs to be pressed, while the key is not stored on disk, not shown on terminal and firmware binary files are already encrypted.

``python esptool\espefuse.py --chip esp32c3 --port COM7 burn_key BLOCK_KEY0 0000111122223333444455556666777788889999aaaabbbbccccddddeeeeffff XTS_AES_128_KEY --text-key --do-not-confirm``

As I only have an ESP32C3, this mod was just tested on it. I put the equivalent code for other ESP models, but they need to be tested though.

### DIFF
--- a/docs/en/espefuse/index.rst
+++ b/docs/en/espefuse/index.rst
@@ -56,6 +56,7 @@ Optional General Arguments Of Commands
 - ``--virt`` - For host tests. The tool will work in the virtual mode (without connecting to a chip).
 - ``--path-efuse-file`` - For host tests. Use it together with ``--virt`` option. The tool will work in the virtual mode (without connecting to a chip) and save eFuse memory to a given file. If the file does not exists the tool creates it. To reset written eFuses just delete the file. Usage: ``--path-efuse-file efuse_memory.bin``.
 - ``--do-not-confirm`` - Do not pause for confirmation before permanently writing eFuses. Use with caution. If this option is not used, a manual confirmation step is required, you need to enter the word ``BURN`` to continue burning.
+- ``--text-key`` - Data input for the ``keyfile`` argument is a 32 character string representing the encryption key itself instead of a string to a file path containing the key. When using this option, key content is also hidden in terminal output while burning fuses. Combining ``--text-key`` and ``--do-not-confirm`` options, esptool can then be invoked in an automated way by another software for prodcution flashing of encrypted firmware. In this case, neither the raw binary of firmware or the encryption key data are exposed outside the development environment.
 
 Virtual mode
 ^^^^^^^^^^^^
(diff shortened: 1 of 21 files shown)


KonstantinKondrashov:
Hi @ysba! From a security perspective, it is not a good approach to support this feature.

You can try to use another way that is already supported. Just use the `burn_efuse` command. [New values can be a numeric value in decimal or hex (with “0x” prefix)](https://docs.espressif.com/projects/esptool/en/latest/esp32/espefuse/burn-efuse-cmd.html#burn-efuse) it gives a different order. Note: the flash encryption key has reversed order in efuse, and the secure boot key has the right order.
```
espefuse.py burn_efuse BLOCK_KEY0  0x123456781234567812345678123456781234567812345678123456781234FFFF
BLOCK_KEY0      (BLOCK4          ) [4 ] read_regs: 1234ffff 12345678 12345678 12345678 12345678 12345678 12345678 12345678

espefuse.py burn_efuse BLOCK_KEY0  123456781234567812345678123456781234567812345678123456781234FFFF
BLOCK_KEY0      (BLOCK4          ) [4 ] read_regs: 78563412 78563412 78563412 78563412 78563412 78563412 78563412 ffff3412
```

Use these flags ` espefuse --debug --virt --path-efuse-file test.bin ...`  to test it and figure out the order of data.

[espefuse supports multiple commands so you can burn efuses at one time.](https://docs.espressif.com/projects/esptool/en/latest/esp32/espefuse/index.html#perform-multiple-operations-in-a-single-espefuse-run).

These commands below do the same as your feature:
1. Burn Key + Key write protection, KEY_PURPOSE_0 + KEY_PURPOSE_0 write protection
```
espefuse --debug --virt --path-efuse-file test.bin -c esp32c3 \
  burn_efuse BLOCK_KEY0 123456781234567812345678123456781234567812345678123456781234FFFF \
                    KEY_PURPOSE_0 XTS_AES_128_KEY  \
  write_protect_efuse BLOCK_KEY0 KEY_PURPOSE_0
```

2. Burn read protection for Key. If it is Flash encryption key (XTS_AES_128_KEY).
```
espefuse --debug --virt --path-efuse-file test.bin -c esp32c3  read_protect_efuse BLOCK_KEY0
```

ysba:
Hi @KonstantinKondrashov! I appreciate the consideration regarding this issue. Indeed, the way you have suggested works just fine for my application. Thank you!

PR #874, opened 2023-04-20
Image info stuff (ESPTOOL-675)

xyzzy42:
# Description of change
Add some additional decoding for the image_info command.

Linux.  image_info doesn't interact with boards.

### DIFF
--- a/esptool/cmds.py
+++ b/esptool/cmds.py
@@ -737,7 +737,10 @@ def image_info(args):
             title = "{} extended image header".format(args.chip.upper())
             print(title)
             print("=" * len(title))
-            print("WP pin: {:#02x}".format(image.wp_pin))
+            print(
+                f"WP pin: {image.wp_pin:#02x}",
+                *["(disabled)"] if image.wp_pin == image.WP_PIN_DISABLED else [],
+            )
             print(
                 "Flash pins drive settings: "
                 "clk_drv: {:#02x}, q_drv: {:#02x}, d_drv: {:#02x}, "
--- a/esptool/cmds.py
+++ b/esptool/cmds.py
@@ -753,7 +753,15 @@ def image_info(args):
                     image.wp_drv,
                 )
             )
-            print("Chip ID: {}".format(image.chip_id))
+            try:
+                chip = next(
+                    chip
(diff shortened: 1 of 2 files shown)


test/test_image_info.py
```diff
@@ -109,7 +109,8 @@ def test_v2_esp32c3(self):
         hdr = read_image("bootloader_esp32c3.bin")[:8]
         ex_hdr = read_image("bootloader_esp32c3.bin")[8:24]
         assert f"Segments: {hdr[1]}" in out, "Wrong num of segments"
-        assert f"WP pin: {ex_hdr[0]:#02x}" in out, "Wrong WP pin"
+        if ex_hdr[0] != 0xee:
```
peterdragun, line 112: Can you please change the hex value to uppercase - `0xEE`? This caused the checks to fail.

esptool/cmds.py
```diff
@@ -737,7 +737,12 @@ def get_key_from_value(dict, val):
             title = "{} extended image header".format(args.chip.upper())
             print(title)
             print("=" * len(title))
-            print("WP pin: {:#02x}".format(image.wp_pin))
+            print(
+                "WP pin:",
+                "disabled"
```
peterdragun, line 742: Also showing the raw value would be nice in some cases, we can keep it in parentheses.

peterdragun (commented):
Thank you for your contribution. Overall the improvement seems good, I left some minor comments. After fixing those we are good to go.

Can you please also rebase the PR?

esptool/cmds.py
xyzzy42: How about:
```
WP pin: 0xee (disabled)
WP pin: 0x12
```
That way the hex value is the same format in both cases.

xyzzy42:
> Can you please also rebase the PR?

It's based off of latest master of this repo.   Is there something else to use?  I don't see a dev branch.

esptool/cmds.py
peterdragun: Perfect!

peterdragun:
You are right, it is okay. Sorry, I somehow overlooked the latest commit and thought you are missing it.

PR #880, opened 2023-05-08
Fix USB-JTAG-Serial PID detection error (ESPTOOL-682)

fuzeman:
# Description of change

Fixes the detection of USB-JTAG-Serial on ports COM10 and greater due to incorrect PID retrieval.

Commit https://github.com/espressif/esptool/commit/00f4bc967c94b25e3190ebc007a0bf132ae42f02 introduced a bug which changed port names to be matched by `in` when not running on darwin, this means `_get_pid` can instead return the PID for `COM1` instead of `COM10` - `COM19`.

### DIFF
--- a/esptool/loader.py
+++ b/esptool/loader.py
@@ -494,12 +494,14 @@ class ESPLoader(object):
         if active_port.startswith("/dev/") and os.path.islink(active_port):
             active_port = os.path.realpath(active_port)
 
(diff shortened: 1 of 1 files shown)


radimkarnis:
Hello @fuzeman,
thanks a lot for the PR and for identifying an issue!

We will put this through our internal review queue and merge ASAP.

PR #900, opened 2023-07-17
espsecure: read_hsm_config now can prompt for HSM PIN (ESPTOOL-716)

rretanubun:
# Description of change

If the config file contains `credentials = prompt` during image signing process, the user will be prompted to type in the HSM PIN. This avoids the need to have the HSM PIN written as plaintext into a config file, which is not a secure practice.

espsecure using NitroKey HSM2 HW on Linux Ubuntu 22.04

### DIFF
--- a/espsecure/esp_hsm_sign/__init__.py
+++ b/espsecure/esp_hsm_sign/__init__.py
@@ -6,6 +6,7 @@ import binascii
 import configparser
 import os
 import sys
+from getpass import getpass
 
 try:
     import pkcs11
@@ -31,10 +32,16 @@ def read_hsm_config(configfile):
     if not config.has_section(section):
         raise configparser.NoSectionError(section)
 
-    section_options = ["pkcs11_lib", "credentials", "slot", "label"]
+    section_options = ["pkcs11_lib", "slot", "label"]
     for option in section_options:
         if not config.has_option(section, option):
             raise configparser.NoOptionError(option, section)
+
+    # If the config file does not contain the "credentials" option,
+    # prompt the user for the HSM PIN
+    if not config.has_option(section, "credentials"):
+        hsm_pin = getpass("Please enter the PIN of your HSM:\n")
+        config.set(section, "credentials", hsm_pin)
 
     return config[section]
 
--- a/docs/en/espsecure/index.rst
+++ b/docs/en/espsecure/index.rst
@@ -50,6 +50,9 @@ HSM config file
 An HSM config file is required with the fields (``pkcs11_lib``, ``credentials``, ``slot``, ``label``, ``label_pubkey``)
 populated corresponding to the HSM used.
 
+To avoid storing the PIN in plaintext, you could simply remove the ``credentials`` option from the config file,
+and during signing using the HSM, you will be prompted to type in the HSM Pin.
+
 Below is a sample HSM config file (``hsm_config.ini``) for using `SoftHSMv2 <https://github.com/opendnssec/SoftHSMv2>`_ as an external HSM: ::
 
     # hsm_config.ini


radimkarnis:
Hello @rretanubun,
thank you for contributing! We will get to this PR soon. If all goes well and it passes our internal review queue, this will get merged.

espsecure/esp_hsm_sign/__init__.py
```diff
@@ -36,6 +37,12 @@ def read_hsm_config(configfile):
         if not config.has_option(section, option):
             raise configparser.NoOptionError(option, section)
 
+    # Check if config.["credentials"] is set to "prompt" and prompt for HSM PIN
+    credentials_config_val = config.get(section, "credentials")
+    if credentials_config_val == "prompt":
+        hsm_pin = getpass("\nPlease enter the PIN of your HSM:\n")
+        config.set(section, "credentials", hsm_pin)
+
```
Harshal5, line 45: Alternatively, rather than using the word "prompt" (might be misleading), we can just remove the `credentials` option from the config file and add a check that if the option is absent, get the credentials using a prompt.

```suggestion
    # If the config file does not contain the "credentials" option, prompt for the HSM PIN
    if not config.has_option(section, "credentials"):
        hsm_pin = getpass("Please enter the PIN of your HSM:\n")
        config.set(section, "credentials", hsm_pin)

```

Harshal5: Could you also add a note in the [documentation](https://github.com/espressif/esptool/blob/master/docs/en/espsecure/index.rst#hsm-config-file) regarding these changes?
Something like,
```
To avoid storing the PIN in plaintext, you could simply remove the `credentials` option from the config file, and during signing using the HSM, you could enter it using the prompt.
```
would suffice.

Harshal5:
Hello @rretanubun,

Thank you for the update!
Just a suggestion regarding the changes.

espsecure/esp_hsm_sign/__init__.py
rretanubun: For documentation update (if we stay with prompt) :  This is what I have in mind:

1. Change the sample to say `credentials = prompt` and then document 
```
Below is a sample HSM config file (hsm_config.ini) for using [SoftHSMv2](https://github.com/opendnssec/SoftHSMv2) as an external HSM:

>> insert config file with credentials = prompt here as in the current example <<

Using this config file, the script will prompt for the HSM pin to be typed in for operations that need HSM access.
Alternatively the keyword `prompt` can be replaced with the `HSM pin` (e.g. 1234) to avoid being prompted.
Note that this is not a good practice for security and using `prompt` is recommended.
```

This way the example (which most people will copy-and paste) already sets them up to the `safer behavior`

Harshal5: @rretanubun There might be a case where the user pin of a token could be the word "prompt" itself (tried using SoftHSM2, it seems possible). In such a situation the user will not be able to get the expected behavior.
Even if say we do not come across such a situation, I think making the `credentials` field optional would be a better approach rather than deciding on a word (say "prompt") and hardcoding it into the tool.

rretanubun: Fair point. Code and documentation updated as requested.

radimkarnis:
Thank you @rretanubun, LGTM! Before merging, could you please:

1) squash the commits
2) fix the issue [detected by the flake8 linter](https://github.com/espressif/esptool/actions/runs/5622423158/job/15235279277?pr=900#step:3:109) (`./espsecure/esp_hsm_sign/__init__.py:40:89: E501 line too long (90 > 88 characters)`)
3) change the commit message to follow the [conventional commits standard](https://www.conventionalcommits.org/en/v1.0.0/#specification). Something like `feat(espsecure): Allow promoting for HSM PIN in read_hsm_config` would be nice. 

Thank you very much!

Issue #888, opened 2023-06-09
Stderr empty when calling esptool with subprocess.Popen (ESPTOOL-696)


Waganama:
### Operating System

Python 

### Esptool Version

V4.6.1

### Python Version

Python 3.11

### Full Esptool Command Line that Was Run

esptool -p COM9 --chip esp32 read_mac

### Esptool Output

```shell
STDOUT : b'esptool.py v4.6.1\r\nSerial port COM9\r\nConnecting......................................\r\n\r\nA fatal error occurred: Failed to connect to ESP32: No serial data received.\r\nFor troubleshooting steps visit:

https://docs.espressif.com/projects/esptool/en/latest/troubleshooting.html\r\n'

STDERR : b''
```

### What is the Expected Behaviour?

STDOUT : b'esptool.py v4.6.1\r\nSerial port COM9\r\nConnecting......................................\r\n\r\n

STDERR : b''A fatal error occurred: Failed to connect to ESP32: No serial data received.\r\nFor troubleshooting steps visit:

https://docs.espressif.com/projects/esptool/en/latest/troubleshooting.html\r\n'"

### More Information

Hello guys,
I tried to catch esptool error in stderr from subprocess.Popen with the following code :

def read_mac(self, port) :
        command = [ESPTOOL_PY, '-p', port, '--chip', 'esp32', 'read_mac']
        send_command(command) 
  

def send_command(command: list[str]) :
    out = subprocess.Popen([sys.executable] + command,  stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=False)
    stdout, stderr = out.communicate()
    
    print(f'STDOUT : {stdout}')
    print(f'STDERR : {stderr}')

Thanks for your help :)

radimkarnis:
Hi @Waganama,
thanks for the ticket!

 It will be implemented as a part of the next major release (v5.0) since it is a breaking change for current scripts relying on reading the stdout output.
 
 I will keep this open until then.

PR #913, opened 2023-08-24
feat: print exceptions to stderr (ESPTOOL-731)

Massimiliano-solutiontech:
Now the exceptions are printed out in the `stderr` instead of `stdout`. That could be a breaking change for some users so here two considerations:

1. make this feature opt-in by a parameter or via `esptool.cfg`
2. the existing scripts can redirect the `stderr` to `stdout` by doing `esptool.py flash_id &> stdout+stderr.txt`

I found this open issue #888 but in my opinion this change can be merged in the current version (`4.7`)

### DIFF
--- a/espefuse/__init__.py
+++ b/espefuse/__init__.py
@@ -294,7 +294,7 @@ def _main():
     try:
         main()
     except esptool.FatalError as e:
-        print("\nA fatal error occurred: %s" % e)
+        print("\nA fatal error occurred: %s" % e, file=sys.stderr)
         sys.exit(2)
 
 
--- a/espefuse/__init__.py
+++ b/espefuse/__init__.py
@@ -20,7 +20,7 @@ import espefuse.efuse.esp32s3 as esp32s3_efuse
 import espefuse.efuse.esp32s3beta2 as esp32s3beta2_efuse
 
 import esptool
-from esptool.loader import REDIRECT_ERROR
+from esptool.loader import REDIRECT_ERRORS
 
 DefChip = namedtuple("DefChip", ["chip_name", "efuse_lib", "chip_class"])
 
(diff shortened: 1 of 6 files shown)


radimkarnis:
Hello @Massimiliano-solutiontech,

Unfortunately, this cannot be merged at this moment, as it is a breaking change. It would require a new major release (v5.0). https://github.com/espressif/esptool/issues/888 is a duplicate and it's planned to be done as a part of the next major release.

> make this feature opt-in by a parameter or via esptool.cfg

This sounds like a reasonable solution in the meantime. Would you be interested in doing that as a part of this PR?

> the existing scripts can redirect the stderr to stdout by doing esptool.py flash_id &> stdout+stderr.txt

Unfortunately, this means it is still a breaking change if the existing scripts have to be changed.

Massimiliano-solutiontech:
> This sounds like a reasonable solution in the meantime. Would you be interested in doing that as a part of this PR?

Yes, in the next few days I'll update this PR

Thanks

Massimiliano-solutiontech:
Hi @radimkarnis,
I did the modifications now the default behavior is restored (errors are printed to `stdout`)

Now you can redirect errors to `stderr` by:

- adding `--redirect-errors` arg (eg. `esptool.py --redirect-errors flash_id`
- adding `redirect_errors = true` to the `esptool.cfg` file

Let me know if the name of the arg fits and/or if the implementation that I did is ok

radimkarnis:
Hello @Massimiliano-solutiontech,

thanks for the implementation!

I will ask you to kindly remove the `--redirect-errors` arg. We try to keep esptool options and arguments as lean as possible. The configuration file is enough to activate this feature until it's made default in the next major release.

Massimiliano-solutiontech:
@radimkarnis done

PR #940, opened 2023-12-14
Remove the shebang from uf2_writer.py (ESPTOOL-781)

befeleme:
This is not a script, so there' no need for it.

I package esptool for Fedora Linux. Our automation has complained that there's a shebang in a file that's not executable. 
Upon inspection it seems so - this file is not intended to be executed directly, hence my proposal to remove the shebang from the file.

NO TESTING


radimkarnis:
Hello @befeleme,
thank you for the contribution and thanks for following the correct styling proposed by our bot!

Changes LGTM, we will merge this as soon as this goes through our internal queue.

PR #950, opened 2024-01-28
fix: Allow the espefuse.py to work when coding scheme == 3 (ESPTOOL-803)

klondi:
When the coding scheme efuse has the value 3, the efuse fields should still be considered to be 256 bits long. Unfortunately, the tool fails miserably with error "A fatal error occurred: Coding scheme (3) not supported"

This patch addresses that by handling the case when coding scheme is 0 in the same way as if coding scheme was 1.

ESP32 with coding scheme == 3

NO TESTING

### DIFF
--- a/espefuse/efuse/base_fields.py
+++ b/espefuse/efuse/base_fields.py
@@ -182,7 +182,7 @@ class EfuseBlockBase(EfuseProtectBase):
 
     def get_block_len(self):
         coding_scheme = self.get_coding_scheme()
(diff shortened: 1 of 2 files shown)


KonstantinKondrashov:
Hi @klondi!
Thanks for your contribution. Yes, I confirm that for the coding scheme = 3 there is no support for now. I will check other places where it can affect, it can be easily added. It relates to the ESP32 chip only.

radimkarnis:
@klondi thank you for the contribution!

PR #960, opened 2024-02-27
Hex2image (ESPTOOL-817)

LeZerb:
This change adds the possibility to convert a hex file to an image file with make_image. It also allows to create images for other targets then ESP8266.

Windows 11, ESP32C3 SuperMini

NO TESTING


dobairoland:
See https://github.com/espressif/esptool/pull/959#issuecomment-2217432503

PR #984, opened 2024-06-21
docs(flashing): Update error when connecting to serial port fails on linux (ESPTOOL-874)

OrhanYigitDurmaz:
This pull request makes more description on linux based systems when esptool cant connect to the serial port defined by the user.

Tested on Arch, Ubuntu 22.04 and 20.04 and Windows 10

NO TESTING

### DIFF
--- a/esptool/loader.py
+++ b/esptool/loader.py
@@ -336,8 +336,9 @@ class ESPLoader(object):
                         [  # permission denied error
                             re.compile(r"Permission denied", re.IGNORECASE),
                             (
-                                "Try to add user into dialout group: "
-                                "sudo usermod -a -G dialout $USER"
+                                "Try to add user into a group that has serial port permissions: "
+                                "Ubuntu and ubuntu-based: sudo usermod -a -G dialout $USER"
+                                "Arch: sudo usermod -a -G uucp $USER"
                             ),
                         ],
                     )


radimkarnis:
Hello @OrhanYigitDurmaz,
thanks for the contribution, it is appreciated!

Adding more instructions is always welcome. In this case, we believe that instead of adding instructions for different Linux distros directly to the error message, we should go the opposite way - simplify the message and extend the [Troubleshooting guide section in the docs](https://docs.espressif.com/projects/esptool/en/latest/esp32/troubleshooting.html) instead. Would you be interested to move the information there?

Also, please take a look at the github-actions bot recommendations above this comment. The commit message is missing the type and scope.

Thanks!

OrhanYigitDurmaz:
> Hello @OrhanYigitDurmaz, thanks for the contribution, it is appreciated!
> 
> Adding more instructions is always welcome. In this case, we believe that instead of adding instructions for different Linux distros directly to the error message, we should go the opposite way - simplify the message and extend the [Troubleshooting guide section in the docs](https://docs.espressif.com/projects/esptool/en/latest/esp32/troubleshooting.html) instead. Would you be interested to move the information there?
> 
> Also, please take a look at the github-actions bot recommendations above this comment. The commit message is missing the type and scope.
> 
> Thanks!

You're right. Then giving the link to [here](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/get-started/establish-serial-connection.html#adding-user-to-dialout-or-uucp-on-linux) is better. Maybe you could also add this to the troubleshooting section?

radimkarnis:
> You're right. Then giving the link to [here](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/get-started/establish-serial-connection.html#adding-user-to-dialout-or-uucp-on-linux) is better. Maybe you could also add this to the troubleshooting section?

Do you want to finish the PR, or should we take over, then?

> Also looks like lint_esptool has some broken dependencies and thus, fails.

That's unrelated to this PR, don't worry about the failure.

Issue #987, opened 2024-07-01
Add ability to specify serial port VID+PID (ESPTOOL-877)


bryghtlabs-richard:
### Is your feature request related to a problem?

Sometimes the computer running esptool has a lot of serial ports. When not specifying a COM# or device file, this can cause esptool to take a long time to find a device, as it searches through every COM# looking for an ESP chip.

### Describe the solution you'd like

I'd like the ability to specify a USB VID/PID (or a few) for filtering possible serial ports for ESPtool to connect to, and it would use this information to skip COM# that do not match.

### Describe alternatives you've considered

It's possible to specify a COM#, but when programming a number of boards with their own COM#, it's tricky to keep track of. Another alternative would be to wrap esptool with something to identify the COM#, then pass to esptool.

PR #988, opened 2024-07-01
feat(esptool): allow filtering ports by VID/PID (ESPTOOL-878)

bryghtlabs-richard:
Enables user to specify --usbVid and/or --usbPid, which filter the com port list before checking for ESP chips. With many serial ports on the computer, this greatly speeds up the process of esptool identifying the correct com port by allowing esptool to skip all com ports without a matching VID/PID. Implements https://github.com/espressif/esptool/issues/987

NO TESTING ON MAC/LINUX. Tested on Windows64 with ESP32-S3-DevKitC and custom ESP32-S3 board.

```
$ time esptool read_flash_status
esptool.py v4.7.0
Found 4 serial ports
Serial port COM30
Connecting......................................
COM30 failed to connect: Failed to connect to Espressif device: No serial data received.
For troubleshooting steps visit: https://docs.espressif.com/projects/esptool/en/latest/troubleshooting.html
Serial port COM29
Connecting......................................
COM29 failed to connect: Failed to connect to Espressif device: No serial data received.
For troubleshooting steps visit: https://docs.espressif.com/projects/esptool/en/latest/troubleshooting.html
Serial port COM18
Connecting...
Detecting chip type... ESP32-S3
...long text omitted...
real    0m18.193s
user    0m0.000s
sys     0m0.015s
```

```
$ time esptool --usbVid 0x303a read_flash_status
esptool.py v4.7.0
Found 1 serial ports
Serial port COM18
Connecting...
Detecting chip type... ESP32-S3
...long text omitted...
real    0m1.067s
user    0m0.000s
sys     0m0.000s
```

$ pytest test_esptool.py --port COM49 --chip esp32s3 --baud 230400
============================= test session starts =============================
platform win32 -- Python 3.11.5, pytest-8.2.2, pluggy-1.5.0
rootdir: C:\Projects\esptool
configfile: pyproject.toml
collected 88 items

test_esptool.py sss.......ss.................s.s.........s....sss.s..... [ 63%]
......s...sssssss..ssss..s..ssss                                         [100%]

============================== warnings summary ===============================
test_esptool.py:120
  C:\Projects\esptool\test\test_esptool.py:120: PytestUnknownMarkWarning: Unknown pytest.mark.flaky - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.flaky(reruns=1, condition=arg_preload_port is not False)

test_esptool.py:1204
  C:\Projects\esptool\test\test_esptool.py:1204: PytestUnknownMarkWarning: Unknown pytest.mark.flaky - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.flaky(reruns=5)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========== 59 passed, 29 skipped, 2 warnings in 415.91s (0:06:55) ============

### DIFF
--- a/docs/en/esptool/advanced-options.rst
+++ b/docs/en/esptool/advanced-options.rst
@@ -117,3 +117,21 @@ An example of this is available in the :ref:`merge_bin <merge-bin>` command desc
 .. note:: PowerShell users
 
     Because of `splatting <https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_splatting?view=powershell-7.3>`__ in PowerShell (method of passing a collection of parameter values to a command as a unit) there is a need to add quotes around @filename.txt ("@filename.txt") to be correctly resolved.
+
+Filtering serial ports
+----------------------
+.. _filtering_serial_ports:
+
+``--port-filter <FilterType>=<FilterValue>`` allows limiting ports that will be tried. This can be useful when esptool is run on a system
+with many serial ports. There are a few different types that can be combined. A port must match all specified FilterTypes, and must match
+at least one FilterValue for each specified FilterType to be considered. Example filter configurations:
+
+.. list::
+
+    * ``--port-filter vid=0x303A`` matches ports with the Espressif USB VID.
+    * ``--port-filter vid=0x303A --port-filter vid=0x0403`` matches Espressif and FTDI ports by VID.
+    * ``--port-filter vid=0x303A --port-filter pid=0x0002`` matches Espressif ESP32-S2 in USB-OTG mode by VID and PID.
+    * ``--port-filter vid=0x303A --port-filter pid=0x1001`` matches Espressif USB-Serial/JTAG unit used by multiple chips by VID and PID.
+    * ``--port-filter name=ttyUSB`` matches ports where the port name contains the specified text.
+
+See also the `Espressif USB customer-allocated PID repository <https://github.com/espressif/usb-pids>`_
--- a/esptool/__init__.py
+++ b/esptool/__init__.py
@@ -124,6 +124,14 @@ def main(argv=None, esp=None):
     )
 
     parser.add_argument(
+        "--port-filter",
+        action="append",
+        help="Serial port device filter, can be vid=NUMBER, pid=NUMBER, name=SUBSTRING",
+        type=str,
+        default=[],
+    )
+
+    parser.add_argument(
         "--before",
         help="What to do before connecting to the chip",
         choices=["default_reset", "usb_reset", "no_reset", "no_reset_no_sync"],
@@ -690,6 +698,23 @@ def main(argv=None, esp=None):
 
     StubFlasher.set_preferred_stub_subdir(args.stub_version)
 
+    # Parse filter arguments into separate lists
+    args.filterVids = []
+    args.filterPids = []
+    args.filterNames = []
+    for f in args.port_filter:
+        kvp = f.split("=")
+        if len(kvp) != 2:
+            raise FatalError("Option --port-filter argument must consist of key=value")
+        if kvp[0] == "vid":
+            args.filterVids.append(arg_auto_int(kvp[1]))
+        elif kvp[0] == "pid":
+            args.filterPids.append(arg_auto_int(kvp[1]))
+        elif kvp[0] == "name":
+            args.filterNames.append(kvp[1])
+        else:
+            raise FatalError("Option --port-filter argument key not recognized")
+
     # operation function can take 1 arg (args), 2 args (esp, arg)
     # or be a member function of the ESPLoader class.
 
@@ -725,7 +750,7 @@ def main(argv=None, esp=None):
             initial_baud = args.baud
 
         if args.port is None:
-            ser_list = get_port_list()
+            ser_list = get_port_list(args.filterVids, args.filterPids, args.filterNames)
             print("Found %d serial ports" % len(ser_list))
         else:
             ser_list = [args.port]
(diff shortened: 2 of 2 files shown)


dobairoland:
Thank you for the contribution. I think this is a good idea but we might think about a little what new CLI arguments we add. I'm worried that others might need other type of filtering (e.g. serial_number, location) for which we would have to add also new arguments. 

- This is one solution to merge the arguments: `--port-filter <vid>:<pid>` where `<vid>` and `<pid>` could be `*` as well. 
- An even more general solution: `--port-filter vid=<number> --port-filter pid=<number>` where `--port-filter` can be repeated several times and can be easily extended.

What do you think @radimkarnis, @peterdragun?

peterdragun:
Overall I like the idea, but as mentioned I would go with one argument for both options.
Maybe we can extend the filter to also cover the port name? For example on macOS ports can have different names based on whether they are using a USB to UART bridge (`/dev/cu.usbserial-XX`) or USB Serial/JTAG (`/dev/cu.usbmodemXX`). So in case I know which peripheral I want to use but don't know the number it may be helpful. I mean we can do that based on VID as well, but it would be easier to remember the port name than VID.

dobairoland:
> Overall I like the idea, but as mentioned I would go with one argument for both options. Maybe we can extend the filter to also cover the port name? For example on macOS ports can have different names based on whether they are using a USB to UART bridge (`/dev/cu.usbserial-XX`) or USB Serial/JTAG (`/dev/cu.usbmodemXX`). So in case I know which peripheral I want to use but don't know the number it may be helpful. I mean we can do that based on VID as well, but it would be easier to remember the port name than VID.

I think this could be covered by the following if we go with my second proposal:
```
--port-filter name=usbserial
```

bryghtlabs-richard:
> An even more general solution: --port-filter vid=<number> --port-filter pid=<number> where --port-filter can be repeated several times and can be easily extended.

I like this.

> I think this could be covered by the following if we go with my second proposal: `--port-filter name=usbserial`

Are you thinking a substring search, or should we check against a prefix from the pyserial port name, like `name=/dev/cu.usbserial`?

dobairoland:
> Are you thinking a substring search, or should we check against a prefix from the pyserial port name, like name=/dev/cu.usbserial?

Substring search is what I had in my mind. `ttyUSB` and `ttyACM` could be easily used on Linux, or even simply `USB` or `ACM`. But this was just a suggestion so we don't end up needing several similar arguments. You don't have to address this in this PR (if you don't want).

On the other hand, what would be great to update in this PR is the documentation and mention this new feature somewhere. Perhaps in a separate section in https://docs.espressif.com/projects/esptool/en/latest/esp32/esptool/advanced-options.html.

dobairoland:
Hi @bryghtlabs-richard. Could you please get back to this pull request and finish it as we've discussed?

bryghtlabs-richard:
Hi @dobairoland , I'm back from a vacation and I think I've got everything we want, and updated the documentation.

docs/en/esptool/advanced-options.rst
```diff
@@ -117,3 +117,18 @@ An example of this is available in the :ref:`merge_bin <merge-bin>` command desc
 .. note:: PowerShell users
 
     Because of `splatting <https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_splatting?view=powershell-7.3>`__ in PowerShell (method of passing a collection of parameter values to a command as a unit) there is a need to add quotes around @filename.txt ("@filename.txt") to be correctly resolved.
+
+Filtering serial ports
+----------------------
+.. _filtering_serial_ports:
+
+``--port-filter <FilterType>=<FilterValue>`` allows limiting ports that will be tried. This can be useful when esptool is run on a system
+with many serial ports There are a few different types that can be combined. A port must match all specified FilterTypes, and must match
```
dobairoland, line 126: "ports There" -> "ports. There"

esptool/__init__.py
```diff
@@ -1010,21 +1035,34 @@ def arg_auto_chunk_size(string: str) -> int:
...
+    for port in list_ports.comports():
+        if sys.platform == "darwin" and port.device.endswith(
+            ("Bluetooth-Incoming-Port", "wlan-debug")
+        ):
+            continue
+        if vids and (port.vid is None or port.vid not in vids):
+            continue
+        if pids and (port.pid is None or port.pid not in pids):
+            continue
+        if names and (port.name is None or not substr_in_list(port.name, names)):
```
dobairoland, line 1062: This substr_in_list is very specific and with inverting the result it is pretty hard to understand.  Could we replace it with an inline check similar to the following?

```python
... or any(name not in port.name for name in names)
```

esptool/__init__.py
```diff
@@ -123,6 +123,14 @@ def main(argv=None, esp=None):
         default=os.environ.get("ESPTOOL_BAUD", ESPLoader.ESP_ROM_BAUD),
     )
 
+    parser.add_argument(
+        "--port-filter",
+        action="append",
+        help="Serial port device filter, can be vid=INT, pid=INT, name=SUBSTRING",
```
dobairoland, line 129: (optional) I think NUMBER instead of INT would be clearer for users.

dobairoland (changes requested):
Thanks. This looks pretty good. I've left a couple of last suggestions.

bryghtlabs-richard:
Thanks Roland, I've updated the commit with your suggestions.

esptool/__init__.py
dobairoland: And I see I've made a mistake. You corrected it to `not any()`. That could be `all()` instead. You can ignore this if you won't make any more changes to this PR.

docs/en/esptool/advanced-options.rst
```diff
@@ -117,3 +117,18 @@ An example of this is available in the :ref:`merge_bin <merge-bin>` command desc
...
+
+``--port-filter <FilterType>=<FilterValue>`` allows limiting ports that will be tried. This can be useful when esptool is run on a system
+with many serial ports. There are a few different types that can be combined. A port must match all specified FilterTypes, and must match
+at least one FilterValue for each specified FilterType to be considered. Example filter configurations:
+
+.. list::
+
+    * ``--port-filter vid=0x303A`` matches ports with the Espressif USB VID.
+    * ``--port-filter vid=0x303A --port-filter vid=0x0403`` matches Espressif and FTDI ports by VID.
+    * ``--port-filter vid=0x303A --port-filter pid=0x1001`` matches Espressif ESP32-S3 ports by VID and PID.
```
radimkarnis, line 133: `303A`:`1001` is not actually just the  ESP32-S3, but the USB-Serial/JTAG unit in general. That means any chip with this peripheral will show up with this VID:PID (S3, C3, C6, H2, P4, ...). 

If you want to mention a specific chip, USB-OTG mode has to be used. e.g. `ESP32-S2` = `VID:PID=303A:0002`

radimkarnis:
Suggestion: A nice touch would be to add a mention about the Espressif USB customer-allocated PID repository
https://github.com/espressif/usb-pids. Essentially we allow users to register a PID under the Espressif VID for a specific piece of hardware. So one of the examples could maybe mention this?

Feel free to ignore, though!

Jason2866:
Agree, nice feature!

dobairoland:
@bryghtlabs-richard Thank you for the contribution. It was merged.

PR #989, opened 2024-07-04
Proposed Edit to Minimize Future Upload Errors (ESPTOOL-879)

JohnGF:
Based on community forums and personal experience, I propose an edit to the documentation/code to help minimize future errors during code uploads. This change aims to clarify that, in at least some implementations of the chip, it is necessary to hold the boot button until the connection is established when attempting to upload code or firmware.

I apologize if this change is deemed unnecessary, but I believe it will help prevent potential issues for users.

Thank you for considering this request.
Best regards,
João GF

### DIFF
--- a/docs/en/advanced-topics/boot-mode-selection.rst
+++ b/docs/en/advanced-topics/boot-mode-selection.rst
@@ -164,6 +164,7 @@ Manual Bootloader
 Depending on the kind of hardware you have, it may also be possible to manually put your {IDF_TARGET_NAME} board into Firmware Download mode (reset).
 
 - For development boards produced by Espressif, this information can be found in the respective getting started guides or user guides. For example, to manually reset a development board, hold down the **Boot** button (``{IDF_TARGET_STRAP_BOOT_GPIO}``) and press the **EN** button (``EN`` (``CHIP_PU``)).
+- Note: Only release **Boot** button when connection is established.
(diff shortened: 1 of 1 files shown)


docs/en/advanced-topics/boot-mode-selection.rst
```diff
@@ -164,6 +164,7 @@ Manual Bootloader
 Depending on the kind of hardware you have, it may also be possible to manually put your {IDF_TARGET_NAME} board into Firmware Download mode (reset).
 
 - For development boards produced by Espressif, this information can be found in the respective getting started guides or user guides. For example, to manually reset a development board, hold down the **Boot** button (``{IDF_TARGET_STRAP_BOOT_GPIO}``) and press the **EN** button (``EN`` (``CHIP_PU``)).
+- Note: Only release **Boot** button when connection is established.
```
dobairoland, line 167: This is not entirely true. It is not needed to enter Download mode, just to stay there if for some reason esptool's reset sequence on connection doesn't work.

I think we could add a note as the last item stating that this in practice means to hold the boot button or pull the boot pin down while starting esptool (with the default reset sequence) and keeping this until esptool establishes connection.

dobairoland (changes requested):
Thank you for your pull request. I've left one suggestion.

Also please see the contribution guide and install the pre-commit hooks which will guide you to make commits in the required format.

Issue #996, opened 2024-07-18
SystemExit Raised Instead of Warning in _check_if_can_reset Method when chip is placed into download mode using GPIO0 (ESPTOOL-893)


DocKenny:
### Operating System

Windows 10

### Esptool Version

esptool-v4.7.0

### Python Version

Python 3.11.7

### Chip Description

ESP32S2 and ESP32S3

### Device Description

ESP32-S2-WROOM
![image](https://github.com/user-attachments/assets/5a25ee5e-2161-443a-9701-b2f68d098f25)

### How is Esptool Run

Arduino IDE

### Esptool Output

```shell
Sketch uses 268886 bytes (20%) of program storage space. Maximum is 1310720 bytes.
Global variables use 14128 bytes (4%) of dynamic memory, leaving 313552 bytes for local variables. Maximum is 327680 bytes.
esptool.py v4.6
Serial port COM4
Connecting....
Chip is ESP32-S2 (revision v0.0)
Features: WiFi, No Embedded Flash, No Embedded PSRAM, ADC and temperature sensor calibration in BLK2 of efuse V2
Crystal is 40MHz
MAC: 70:04:1d:52:dc:dc
Uploading stub...
Running stub...
Stub running...
Changing baud rate to 921600
Changed.
Configuring flash size...
Flash will be erased from 0x00001000 to 0x00004fff...
Flash will be erased from 0x00008000 to 0x00008fff...
Flash will be erased from 0x0000e000 to 0x0000ffff...
Flash will be erased from 0x00010000 to 0x00051fff...
Compressed 15568 bytes to 11011...
Writing at 0x00001000... (16 %)
Writing at 0x00001c23... (33 %)
Writing at 0x00002767... (50 %)
Writing at 0x00003244... (66 %)
Writing at 0x00003ba6... (83 %)
Writing at 0x00004887... (100 %)
Wrote 15568 bytes (11011 compressed) at 0x00001000 in 0.5 seconds (effective 264.8 kbit/s)...
Hash of data verified.
Compressed 3072 bytes to 146...
Writing at 0x00008000... (100 %)
Wrote 3072 bytes (146 compressed) at 0x00008000 in 0.1 seconds (effective 215.1 kbit/s)...
Hash of data verified.
Compressed 8192 bytes to 47...
Writing at 0x0000e000... (100 %)
Wrote 8192 bytes (47 compressed) at 0x0000e000 in 0.1 seconds (effective 519.6 kbit/s)...
Hash of data verified.
[...]

Issue #912, opened 2023-08-23
Return code 1 for the " can not exit the download mode over USB" warning confuses external processes


EternityForest:
### Is your feature request related to a problem?

When a manual reset is required, the flasher process exits with return code 1, causing scripts that call it to think there was a failure.

### Describe the solution you'd like

I think it would be helpful to either change it to return 0 unless there is a true error, or better yet reserve and document a specific error code just for indicating that a manual reset is required, so that automated tools can correctly interpret things.

I'm not that familiar with the tool in general, but I think having some well defined way to access the status would be helpful.

radimkarnis:
Hello @EternityForest,

the reset behavior can be changed with the `--after` option. If `--after no_reset` is used, then the error message is not printed and esptool exits with return code 0.

`--after hard_reset` is set by default, resulting in the error if USB-OTG mode is used (other modes don't require manual reset).

I believe the correct fix is for the automated tools to correctly configure esptool and set the `--after` option accordingly if USB-OTG is used to prevent the failure.

EternityForest:
It would be nice to keep the message so the user knows what to do, and still try to reset if it's possible to do so.  I suppose you could keep a list on the tool side of which boards should get which option, but that seems not very DRY.

Maybe an `--after auto_reset` option could be added to do it?

PR #997, opened 2024-07-18
Fix Warning Handling in `_check_if_can_reset` Method (ESPTOOL-894)

DocKenny:
### Description of Change
This change modifies the `_check_if_can_reset` method to issue a warning instead of raising a `SystemExit` when the chip is in download mode using GPIO0. This ensures that the method behaves as intended by only issuing a warning without terminating the program.

### Bug Fixes
This change fixes the following bug:
- [#996](https://github.com/espressif/esptool/issues/996)

### Hardware & Software Combinations Tested
I have tested this change with the following hardware & software combinations:
- **Operating System**: Windows 10
- **Development Board**: ESP32-S2
- **Baud Rate**: 115200

### Automated Integration Tests
I have run the `esptool.py` automated integration tests with this change and the above hardware:
- **Results**: 60 tests passed, 29 tests skipped

```sh
pytest test_esptool.py --port COM4 --chip esp32s2 --baud 115200

### DIFF
--- a/esptool/targets/esp32s2.py
+++ b/esptool/targets/esp32s2.py
@@ -5,6 +5,7 @@
 
 import os
 import struct
+import warnings
 from typing import Dict
 
 from .esp32 import ESP32ROM
@@ -296,15 +297,15 @@ class ESP32S2ROM(ESP32ROM):
             strap_reg & self.GPIO_STRAP_SPI_BOOT_MASK == 0
             and force_dl_reg & self.RTC_CNTL_FORCE_DOWNLOAD_BOOT_MASK == 0
         ):
-            print(
+            warnings.warn(
                 "WARNING: {} chip was placed into download mode using GPIO0.\n"
                 "esptool.py can not exit the download mode over USB. "
                 "To run the app, reset the chip manually.\n"
                 "To suppress this note, set --after option to 'no_reset'.".format(
                     self.get_chip_description()
-                )
+                ),
+                UserWarning
             )
-            raise SystemExit(1)
 
     def hard_reset(self):
         uses_usb_otg = self.uses_usb_otg()
--- a/esptool/targets/esp32s3.py
+++ b/esptool/targets/esp32s3.py
@@ -5,6 +5,7 @@
 
 import os
 import struct
+import warnings
 from typing import Dict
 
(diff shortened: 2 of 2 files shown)


radimkarnis:
Hello @DocKenny,
thank you for your contribution! It is always appreciated. I will communicate here, not in the linked issue. 

We believe the current behavior is correct. Please see this issue with an explanation: https://github.com/espressif/esptool/issues/912.

Long story short, it is up to the user to configure the `--after` argument correctly. If a reset is attempted, but cannot be performed, an error is appropriate. Setting `--after no_reset` in this case will get rid of the error and issue a `0` exit status.

DocKenny:
Hi @radimkarnis 
Thank you for reviewing my pull request.

I wanted to share my perspective on the current implementation. From my testing and experience, even though the code gets successfully loaded onto the chip, the raised SystemExit gives the impression that the program has terminated unexpectedly. This can be misleading to users, especially those who may not be familiar with the underlying behavior of the method.

The error message "Failed uploading: uploading error: exit status 1" may give users the impression that the code was not successfully uploaded to the chip, even though it has been.

Thank you for considering this feedback. I am open to further discussions and suggestions on how we can enhance this feature.

radimkarnis:
Thank you @DocKenny for the feedback!

The `Failed uploading: uploading error: exit status 1` is output from Arduino, not esptool (yes, still, this is caused by the `1` exit status from esptool).

I believe this should be fixed on the Arduino side: If ESP32-S2 or ESP32-S3 (or any other future USB-OTG-enabled chip) is selected in Arduino and USB-OTG is used to download the code, Arduino should set the `--after no_reset` option under the hood to suppress this warning. Correctly configuring the tools should be the job of the SDK.

me-no-dev:
> I believe this should be fixed on the Arduino side: If ESP32-S2 or ESP32-S3 (or any other future USB-OTG-enabled chip) is selected in Arduino and USB-OTG is used to download the code, Arduino should set the --after no_reset option under the hood to suppress this warning.

This warning shows when the board has been put into download mode manually (pressing BOOT and EN). In Arduino we can not detect that  the user has manually put the board in Download mode to change flags. IMHO the warning is fine

PR #1008, opened 2024-09-12
Add magic value for P4 (ESPTOOL-921)

mutatrum:
The P4 on my dev board reports 0xFFFFFFFF as magic value. It's an engineering sample and it's self reporting as silicon version v0.1. I can't really judge if this should be merged though.


radimkarnis:
Hi @mutatrum,
thanks for taking the time to report this, and even to send a PR!

We have recently found out that the magic number autodetection method is not reliable for ESP32-P4, as the value at the arbitrary address (`CHIP_DETECT_MAGIC_REG_ADDR`) can change between runs. Due to this, we are now considering changing the address (for all of the previous targets as well).

Fortunately, this is shouldn't be a big deal for P4, as autodetection based on the magic number is just a fallback method. The `get_security_info` method is primarily used and reliable.

radimkarnis:
We will keep this issue open as a tracker until it's resolved. 

Thanks again.

Issue #1010, opened 2024-09-19
[4.8.0] Breaks esptool import due to esptool.py in Scripts instead of esptool.exe (ESPTOOL-924)


matthuisman:
### Operating System

Windows

### Esptool Version

4.8.0

### Python Version

3.10

### Esptool Output

```shell
from esptool import get_default_connected_device, ESPLoader, DEFAULT_CONNECT_ATTEMPTS
ImportError: cannot import name 'get_default_connected_device' from 'esptool' (\\?\C:\Users\matt\Envs\py310\Scripts\esptool.py)
```

### What is the Expected Behaviour?

Can import from esp module and use esptool.exe cmdline

### More Information

setuptools 75.1.0
pip 24.2
wheel  0.44.0

### Other Steps to Reproduce

the Scripts directory has to be before site-pages in sys.path
virtualenvs do that
![image](https://github.com/user-attachments/assets/d352a190-9ba9-4cf5-ba43-a965127d8293)

matthuisman:
we are running another exe in Scripts so suspect Scripts is inserted to sys.path

matthuisman:
ok. so its just python behaviour to add the cwd first to sys.path
So when our other console script in Scripts runs, python is inserting the Scripts dir to sys.path
And due to esptool.py having same name as the module - its finding the cmdline script

Can it be reverted back to exe?

Also, for users who have setup .py files to open in editor by default, just typing esptool or esptool.py opens the file for editing :(
Need to do python esptool.py ...  now

radimkarnis:
Hi @matthuisman,

I am confused a little by your report. You have several options how to run esptool:
- We recommend [installing it with `pip`](https://docs.espressif.com/projects/esptool/en/latest/installation.html) as a python package. 
- Or you can download an executable binary for Windows (`esptool.exe`) [at the bottom of the release](https://github.com/espressif/esptool/releases/tag/v4.8.0). In this case, you don't need Python or do any installation.

Your report shows you've tried to install it as a Python package, but you keep mentioning `esptool.exe`. 

> Can it be reverted back to exe?

We haven't changed anything in this release and I don't fully understand what you mean by this. What version did you use before `v4.8.0`?

> Also, for users who have setup .py files to open in editor by default, just typing esptool or esptool.py opens the file for editing

Try running simply `esptool` (not `esptool.py`) on windows.

matthuisman:
It has changed. pip install esptool now gives you esptool.py in Scripts
[...]

PR #1011, opened 2024-09-23
ci(pypi_release): Use build --sdist argument to avoid building non-universal whl (ESPTOOL-926)

matthuisman:
https://github.com/espressif/esptool/commit/eacc94a386267ad39de876f29b7bd1a6823387ae changed from using 
python setup.py sdist 
to 
python -m build

The latter now builds a .whl which got uploaded using twine on the 4.8.0 release (https://pypi.org/project/esptool/4.8.0/#files)

Unfortuatnely due to setup.py (https://github.com/espressif/esptool/blob/master/setup.py),
the whl that is built on ubuntu uses scripts instead of entry_points
This makes the whl incompatible with Windows systems and causes import error / no exe's created
See https://github.com/espressif/esptool/issues/1010

This change adds --sdist argument to build which tells it not to create the whl.
This allows all systems to just use the source and then do the correct setup.py behaviour

https://github.com/espressif/esptool/issues/1010

I have tested build --sdist locally in Ubuntu and then installing the results .tar.gz in both windows and ubuntu

NO TESTING

### DIFF
--- a/.github/workflows/dev_release_esptool_pypi.yml
+++ b/.github/workflows/dev_release_esptool_pypi.yml
@@ -36,5 +36,5 @@ jobs:
         python -m pip download esptool==$(python setup.py -V) && echo "Version ${{ github.ref_name }} already published, skipping..." && exit 1
 
         echo "Packaging and publishing new esptool development release: ${{ github.ref_name }}"
-        python -m build
+        python -m build --sdist
         twine upload dist/*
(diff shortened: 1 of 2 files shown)


matthuisman:
the other option would be to do twine upload dist/*.tar.gz instead but this felt cleaner

matthuisman:
@radimkarnis thanks for quick merge

I'd suggest yanking the previous release if possible as I suspect anyone using --prefer-binary will keep getting that broken whl.

PR #1012, opened 2024-09-25
[packaging] Exclude also the submodules of test and docs (ESPTOOL-927)

befeleme:
When running the wheel build of esptool 4.8.0 in Fedora, the `docs/en/conf.py` and `test/efuse_scripts/*.py` were included in the distribution.

Apparently, setuptools excludes only the first level of modules, unless specified otherwise. Following the official documentation, the `Important` part, adding asterisks to exclude all the found submodules does the trick
https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#finding-simple-packages

I've tested it during the Fedora RPM build and verified this change works for us.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,7 +69,7 @@
     "*" = ["esptool/targets/stub_flasher/1/*", "esptool/targets/stub_flasher/2/*"]
 
 [tool.setuptools.packages]
-    find = {exclude = ["ci", "flasher_stub", "test", "docs"]}
+    find = {exclude = ["ci*", "test*", "docs*"]}
 
 [tool.setuptools.dynamic]
     version = {attr = "esptool.__init__.__version__"}


pyproject.toml
```diff
@@ -69,7 +69,7 @@
     "*" = ["esptool/targets/stub_flasher/1/*", "esptool/targets/stub_flasher/2/*"]
 
 [tool.setuptools.packages]
-    find = {exclude = ["ci", "flasher_stub", "test", "docs"]}
+    find = {exclude = ["ci", "flasher_stub", "test*", "docs*"]}
```
radimkarnis, line 72: Hello @befeleme,

The `flasher_stub` is no longer a part of the repository. Could you please also remove it here as a part of this PR?

radimkarnis:
Also, please see the bot response above and change your commit message accordingly.

pyproject.toml
radimkarnis: Also, I wonder - the `ci` directory also contains a `.py` file. That doesn't get included?

befeleme: It doesn't because `ci` is explicitly excluded. If there was a subfolder with `*.py` files, they would get into the distribution. But to make it safe, I can add the * there too.

radimkarnis: I understand, thanks. Adding `*` to `ci` as well sounds good to me.

Issue #178, opened 2017-02-14
Support make_image subcommand for esp32 (ESPTOOL-128)


young-mu:
Now **make_image** subcommand is just for esp8266.

projectgus:
This is correct. elf2image works on both architectures, but make_image only works on ESP8266.

make_image is fairly limited command, and even on ESP8266 the images it generates are not always very useful. I'd recommend generating the ELF file you need (via some combination of linker arguments and objcopy commands) and then using elf2image to convert this to a binary.

Is there a particular use case you have in mind that requires make_image?

young-mu:
OK, this is my scenario. I want to make a binary file to an image and map it to an specified address (like 0x3FFE0000). So in application, I can access these data of binary file by address 0x3FFE0000.

projectgus:
Sorry I didn't write back to you before, I thought I had.

If you're using ESP-IDF then one other way to make this work is to use the "embedding" feature in the IDF build system:
http://esp-idf.readthedocs.io/en/latest/build-system.html#embedding-binary-data

This will allow you to embed the file into the ELF, it'll be mapped to flash cache in .rodata, and you can access it directly as a large read-only buffer in read-only memory.

If you're using a different build system then you can still use the same approach as esp-idf does, using objcopy to embed the file.

I'll keep this feature request open in order to expand make_image, although because the ESP32 has a more complex flash cache it may be harder to implement a useful command.

young-mu:
Ok, thank you. Now I use *spi_flash_mmap* API to achieve my goal.

radimkarnis:
I’m closing this issue because it has been inactive for a few years and it seems like it's not relevant anymore. Please reopen if further discussion is needed.

PR #1024, opened 2024-10-24
make_image: support ESP32 image format (ESPTOOL-946)

sven337:
This change adds ability to generate ESP32 firmware images with make_image.
This had been requested a long time ago and not implemented, but I found use for that feature.

https://github.com/espressif/esptool/issues/178

ESP32-WROOM-32D firmware: dumped, extracted, rebuilt with make_image.
This change will not preserve the flash flags, but these can be manually added.

NO TESTING

### DIFF
--- a/esptool/cmds.py
+++ b/esptool/cmds.py
@@ -20,6 +20,7 @@ from .bin_image import (
     ESP8266ROMFirmwareImage,
     ESP8266V2FirmwareImage,
     ESP8266V3FirmwareImage,
+    ESP32FirmwareImage
 )
 from .loader import (
     DEFAULT_CONNECT_ATTEMPTS,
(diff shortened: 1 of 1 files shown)


dobairoland:
Please see the note at https://docs.espressif.com/projects/esptool/en/latest/esp8266/esptool/advanced-commands.html#assemble-a-firmware-image-make-image:

> In general, it is better to create an ELF image (including any binary data as part of the ELF, by using objcopy or other tools) and then use elf2image to generate the .bin file.

`make_image` was only supported for ESP8266 and it was deprecated after that in favor of `elf2image`. This command will be probably removed in the next major release of esptool.

I'm sorry but we will not accept this PR.

sven337:
Please do not remove it, this is a useful feature.
This patch *trivially* adds support for ESP32, I don't understand what could justify refusing it.
elf2image requires an ELF file but you do not always have that if you're editing segments directly.

Issue #1040, opened 2024-12-07
Docs state the wrong GPIO pins needed for QIO SPI flash mode


michaesc:
### Operating System

Ubuntu 24.04.1 LTS

### Esptool Version

4.8.1

### Python Version

3.12.3

### Chip Description

ESP32-C3

### How is Esptool Run

Bourne shell

### I Have Read the Troubleshooting Guide

- [X] I confirm I have read the troubleshooting guide.

PR #1041, opened 2024-12-07
Correction of extra GPIO pins needed for QIO SPI flash mode (ESPTOOL-972)

michaesc:
# Bug fix
 esptool-1040 (Github bug #1040

# Testing

To test this correction please run the Sphynx documentation generator with the new (pull request) sources.


radimkarnis:
Hello @michaesc,
thanks for contributing! 

I believe your correction only applies to ESP32-C3, right? Currently, your commit will change the pin number in the docs for all the chips. The numbers should change depending on the currently selected chip target.

If you want to see how to do it, search for e.g. `IDF_TARGET_BOOTLOADER_OFFSET` on the same page.

Also, please change the commit name according to the message from the bot above. Adding `docs(flash_modes):` at the beginning should suffice.

PR #1061, opened 2025-01-30
docs: Correct instruction to run (ESPTOOL-1003)

akym:
Previous command `python esptool.py` would only work if the repo is cloned, with instruction above in the documentation to install the pip package, it makes more sense to invoke the module and run the tool as follows.

```
python -m esptool -p PORT flash_id
```

Make instructions clearer and easier to understand.

NO TESTING

NO TESTING

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -29,11 +29,11 @@ Getting started is easy:
 
 2) Connect an Espressif chip to your computer.
 
-3) Run ``esptool.py`` commands. For example, to read information about your chip's SPI flash, run:
+3) Invoke ``esptool`` module with commands. For example, to read information about your chip's SPI flash, run:
 
     ::
 
-        $ esptool.py -p PORT flash_id
+        $ python -m esptool -p PORT flash_id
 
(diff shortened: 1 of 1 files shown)


radimkarnis:
Hi @akym,
thanks for the contribution!

> Previous command python esptool.py would only work if the repo is cloned

1) That's true, if the repo was cloned and esptool is invoked from the sources without installation, `python esptool.py <cmd>` would be the way (targeting [this wrapper file](https://github.com/espressif/esptool/blob/master/esptool.py)). But the command in docs does not have the `python` part - there is a difference between `python esptool.py <cmd>` and `esptool.py <cmd>`.
2) The docs mention the command simply as `esptool.py <cmd>` (without `python`), which is what you can do when installing with `pip`. It creates an alias during installation allowing you to run the tool without having to invoke it as a module (`python -m ...`)

Although the changes you propose would work as well, I don't believe they make things clearer and easier to understand for basic users (or people inexperienced with Python). The quick start guide should be as simple as possible. I am sorry, but for these reasons, I have to close this PR.

Issue #1063, opened 2025-02-05
sys.stdout may not be available in certain contexts (eg. pyinstaller headless, no stdout context) (ESPTOOL-1007)


brentpicasso:
fix: guard against possible None sys.stdout

### Operating System

Reproduces on windows

### Esptool Version

4.8.1 and later

### Python Version

3.11.x

### Full Esptool Command Line that Was Run

N/A

### Esptool Output

```shell
We are using esptool to flash firmware in the context of our app. 

This app runs on Kivy, and built with Pyinstaller. On windows, it appears that sys.stdout is not available (sys.stdout is None)

Under these conditions, esptool will crash when sys.stdout.flush() is called
```

### What is the Expected Behaviour?

esptool should guard against this condition and not crash if sys.stdout is not available.

### More Information

Reproduces on Windows with Kivy app built with pyinstaller

PR #1064, opened 2025-02-05
encapsulate sys.stdout.flush() into logger, guard against None sys.stdout (ESPTOOL-1008)

brentpicasso:
fix(esptool/all) : Encapsulates sys.stdout.flush() in logger, with a guard if stdout is not valid (e.g. None)

Resolves https://github.com/espressif/esptool/issues/1063

We are using esptool to flash firmware in the context of our app. 

This app runs on Kivy, and built with Pyinstaller. On windows, it appears that sys.stdout is not available (sys.stdout is None)

Under these conditions, esptool will crash when sys.stdout.flush() is called

What is the Expected Behaviour?

esptool should guard against this condition and not crash if sys.stdout is not available.

Encapsulates sys.stdout.flush() in logger, with a guard if stdout is not valid (e.g. None)

### DIFF
--- a/esptool/cmds.py
+++ b/esptool/cmds.py
@@ -140,7 +140,7 @@ def detect_chip(
                     connect_mode, connect_attempts, detecting=True, warnings=False
                 )
             log.print("Detecting chip type...", end="")
-            sys.stdout.flush()
+            log.flush()
             chip_magic_value = detect_port.read_reg(
                 ESPLoader.CHIP_DETECT_MAGIC_REG_ADDR
             )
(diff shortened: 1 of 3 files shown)


radimkarnis:
Hi @brentpicasso,
thanks for the report and the PR!

The usage of `sys.stdout.flush()` seems like a remnant from the days of Python 2 support. In Python 3, the default string ending when `print` is called is `"\n"`, which triggers a flush. In other cases (when `end=""` is specified), `print()` can be called with the `flush=True` argument to force a flush. 

That means we can get rid of some of these calls in esptool completely (e.g. every `sys.stdout.flush()` after `print_overwrite()`, because [it calls](https://github.com/espressif/esptool/blob/master/esptool/logger.py#L120) `print(flush=True)`). And in other cases, we can just add the `flush=True` argument to the previous print statements. This way, we don't have to extend the logger class with a new method and proper flushing will be handled automatically.

Would you want to look into this, or do you want me to take over? It would be great if you could at least verify this works in your scenario. Thanks a lot!

brentpicasso:
Thank you @radimkarnis for handling it!

PR #1074, opened 2025-03-19
fix: Close the data file after reading the data (ESPTOOL-1029)

stevanStevic:
I'm building a custom flash data tool for my specific needs. The flasher takes the release zip, which is packaged with a custom python script. If the secure boot is enabled, this release will also contain the digest of the private key, which needs to be flashed.

Relevant steps of the flasher, for this fix, are the following:
- Flasher unzips the release into temporary directory
- Finds the digest file
- Invokes espefuse with key digest to burn it.
- Remove temporary dir 

The first 3 operations are successful
```bash
espefuse.py v4.8.1
Connecting....
Detecting chip type... ESP32-S3

=== Run "burn_key" command ===
Sensitive data will be hidden (see --show-sensitive-info)
Burn keys to blocks:
 - BLOCK_KEY0 -> [?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ??]
        The same value for BLOCK_KEY0 is already burned. Do not change the efuse.
        'KEY_PURPOSE_0' is already 'SECURE_BOOT_DIGEST0'.
        Disabling write to key block
        The same value for WR_DIS is already burned. Do not change the efuse.


Check all blocks for burn...
idx, BLOCK_NAME,          Conclusion
Nothing to burn, see messages above.
Successful
```

but when I want to delete the temporary directory, the operation fails with the following message:

```
[WinError 32] The process can not access the file because it is being used by another process: 'C:\\Users\\Me\\AppData\\Local\\Temp\\tmpvlylxsp5\\digest.bin'
```

I found out that if I close the opened key file, the following operation is successful. I added changes for all other chips, but these are **not tested** on my end.

I found this bug and tested the change (for **esp32s3)** with the following setup:
- Windows 11
- Hardware: https://github.com/rtek1000/YD-ESP32-23 (esp32s3)
- esptool v4.8.1
- Python script
```Python
import os
import shutil
import tempfile
import zipfile
from espefuse import main as espefuse_main


def main():

    # Configuration
    zip_file_path = (
        "C:\\Users\\Me\\Downloads\\test.zip"  # Change this to the actual path
    )
    baud_rate = 460800
    port = "COM6"
    block_idx = 0

    # Step 1: Extract ZIP to a temporary directory
    temp_dir = tempfile.mkdtemp()
    try:
        with zipfile.ZipFile(zip_file_path, "r") as zip_ref:
            zip_ref.extractall(temp_dir)

        # Step 2: Load digest.bin
        digest_file = os.path.join(temp_dir, "digest.bin")
        if not os.path.exists(digest_file):
            raise FileNotFoundError(
                f"Error: {digest_file} not found in extracted files."
            )

        # Step 3: Burn the secure boot public key digest
        espefuse_main(
            [
                "--do-not-confirm",
                "--chip",
                "auto",
                "--port",
                str(port),
                "--baud",
                str(baud_rate),
                "burn_key",
                f"BLOCK_KEY{block_idx}",
                digest_file,
                "SECURE_BOOT_DIGEST0",
            ]
        )
        print("Secure boot eFuse burned with public key digest.")

    finally:
        # Step 4: Cleanup - Remove the temporary directory
        shutil.rmtree(temp_dir)
        print("Temporary files removed.")


if __name__ == "__main__":
    sys.exit(main())

```

NO_TESTING

### DIFF
--- a/espefuse/efuse/esp32/operations.py
+++ b/espefuse/efuse/esp32/operations.py
@@ -229,6 +229,7 @@ def burn_key(esp, efuses, args):
             raise esptool.FatalError("Unknown block name - %s" % (block_name))
         num_bytes = efuse.bit_len // 8
         data = datafile.read()
+        datafile.close()
         revers_msg = None
         if block_name in ("flash_encryption", "secure_boot_v1"):
             revers_msg = "\tReversing the byte order"
--- a/espefuse/efuse/esp32c2/operations.py
+++ b/espefuse/efuse/esp32c2/operations.py
@@ -207,7 +207,11 @@ def burn_key(esp, efuses, args, digest=None):
 
     print("Burn keys to blocks:")
     for datafile, keypurpose in zip(datafile_list, keypurpose_list):
-        data = datafile if isinstance(datafile, bytes) else datafile.read()
+        if isinstance(datafile, bytes):
+            data = datafile
+        else:
+            data = datafile.read()
+            datafile.close()
 
         if keypurpose == "XTS_AES_128_KEY_DERIVED_FROM_128_EFUSE_BITS":
             efuse = efuses["BLOCK_KEY0_LOW_128"]
--- a/espefuse/efuse/esp32c3/operations.py
+++ b/espefuse/efuse/esp32c3/operations.py
@@ -251,6 +251,7 @@ def burn_key(esp, efuses, args, digest=None):
 
         if digest is None:
             data = datafile.read()
+            datafile.close()
         else:
             data = datafile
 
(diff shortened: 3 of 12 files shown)


radimkarnis:
Hi @stevanStevic,
Thank you for the nice PR! 

We are currently trying to improve the usability of esptool, espefuse, and espsecure as Python modules, so this is helpful.

The CI results look good, I will put this through our internal review queue and we'll get this merged soon if possible.

radimkarnis:
Merged, thanks for the contribution!

PR #1096, opened 2025-06-23
progress bar in one line is not working using VSC / Platformio (ESPTOOL-1087)

Jason2866:
Using esptool with Platformio espressif32 together with VSC as IDE on Platform MacOS (Darwin) and Windows

In logger.py the `is_tty = hasattr(sys.stdout, "isatty") and sys.stdout.isatty()` does not work in this setup.
Changed the if to use `term_supports_color` as decider. This works correctly and should be compatible in other setups.
Furthermore the ANSI Escape sequence to position the Cursor was not working. An sequence is added to go back one line to pos1

@radimkarnis

### DIFF
--- a/esptool/logger.py
+++ b/esptool/logger.py
@@ -73,7 +73,6 @@ class EsptoolLogger(TemplateLogger):
     ansi_clear: str = ""
     ansi_line_up: str = ""
     ansi_line_clear: str = ""
-    ansi_line_up_pos1: str = ""
 
     _stage_active: bool = False
     _newline_count: int = 0
@@ -135,7 +134,6 @@ class EsptoolLogger(TemplateLogger):
             cls.instance.ansi_clear = "\033[K"
             cls.instance.ansi_line_up = "\033[1A"
             cls.instance.ansi_line_clear = "\x1b[2K"
-            cls.instance.ansi_line_up_pos1 = "\033[F"
         else:
             cls.instance.ansi_red = ""
             cls.instance.ansi_yellow = ""
@@ -144,7 +142,6 @@ class EsptoolLogger(TemplateLogger):
             cls.instance.ansi_clear = ""
             cls.instance.ansi_line_up = ""
             cls.instance.ansi_line_clear = ""
-            cls.instance.ansi_line_up_pos1 = ""
 
     def print(self, *args, **kwargs):
         """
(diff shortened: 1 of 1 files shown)


radimkarnis:
Hi @Jason2866,
Thanks for the report and for taking the time to contribute!

I am not entirely sure about the `smart_features = is_tty or term_supports_color and not no_color` logic. What if there is a case where `is_tty` is False (ANSI sequences are not supported), but `term_supports_color` is True? 

Also, the revised `pos1` sequence doesn't seem to work as expected in my environment. It starts overwriting the previous log line by line, instead of staying on the same line. There should probably be additional logic we are not thinking of. Can you please try your fix in a different env?

Jason2866:
Well, that's Pandoras Box. Yes, that's what happens. `is_tty` is False and `term_supports_color` is True. The change is tested with Windows and MacOS with VSC using [pioarduino](https://github.com/pioarduino/platform-espressif32) 
The Escape sequence is there working, and the current implemented does not work. With the change (Changed the progress bar there too, not in this PR).

Bildschirmfoto 2025-06-25 um 21 29 11

radimkarnis:
We tried to reproduce this in many environments (MacOS, Linux, and Windows).

1) We can accept the `smart_features = is_tty or term_supports_color and not no_color` change, as this seems to cover more cases 

2) The `pos1` fix, on the other hand, doesn't make much sense. We didn't manage to reproduce this and the fix causes troubles like this (lines being printed upwards, eating the previous log):

image

The logic in the code is - print a new line after the progress bar iteration only if:
- The progress bar is complete
- Smart features are off (no overwriting)

None of these conditions are met in your case, and it is still printing a new line (even when the end char is `""` and not `\n`), how?

Anyway, until this gets figured out, I think it's much more acceptable for the progress bar overwriting to simply not work in some edge cases instead of risking the mentioned upwards-eating scenario-

Jason2866:
> None of these conditions are met in your case, and it is still printing a new line (even when the end char is "" and not \n), how?

How, honestly idk. It is like that. Tested and it needs replacing the "" with the "ansi_line_up_pos1" ANSI sequence.

> Anyway, until this gets figured out, I think it's much more acceptable for the progress bar overwriting to simply not work in some edge cases instead of risking the mentioned upwards-eating scenario-

Fully agree, will use my forked version for pioarduino to have a working progress bar with VSC. Btw. not an edge case. pioarduino is widely used for Arduino espressif projects. I think more than 20% are using pioarduino for espressif Arduino projects. Anyway user does not have to care about. The pioarduino setup routine takes care to install the forked esptool.py.

radimkarnis:
@Jason2866 Thanks for your insight.

To finish this PR, can we limit this to the `is_tty or term_supports_color and not no_color` only?

If you squash and rebase, we can then merge this. Thank you!

Jason2866:
@radimkarnis  changed to only `is_tty or term_supports_color and not no_color` which fixes detection of color support with VSC

PR #1104, opened 2025-07-21
Fix running with click-8.2.1

triaxx:
The pkgsrc framework provides devel/py-click in version 8.2.1. The dependency "click<8.2.0" cannot be respected in this framework. Therefore esptool must be patched to update it to its last release.

### DIFF
--- a/esptool/cli_util.py
+++ b/esptool/cli_util.py
@@ -155,7 +155,7 @@ class AddrFilenamePairType(click.Path):
 
     name = "addr-filename-pair"
 
-    def get_metavar(self, param):
+    def get_metavar(self, param, ctx=None):
         return "<address> <filename>"
 
     def convert(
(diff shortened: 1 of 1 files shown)


esptool/cli_util.py
```diff
@@ -294,7 +294,7 @@ def __init__(self, *args, **kwargs):
         self._eat_all_parser = None
         # Set the metavar dynamically based on the type's metavar
         if self.type and hasattr(self.type, "name"):
-            self.metavar = f"[{self.type.get_metavar(None) or self.type.name.upper()}]"
+            self.metavar = f"[{self.type.get_metavar(None, None) or self.type.name.upper()}]"
```
peterdragun, line 297: This fix makes it work with click >= 8.2.0, but it does not work with older versions (see the failed tests). We would like to keep the compatibility. We should find a solution that works for both versions, specifically <8.2.0 and >=8.2.0. You can take inspiration here: https://github.com/sirosen/click/blob/55189e3ccf6677f3f350d4a1180c7c95142d79d7/docs/support-multiple-versions.md

PR #1108, opened 2025-08-05
fix: stop exit 0 when being called programmatically

hfudev:
The current `esptool.main(...)` call causes the Python process to exit immediately.

This PR bypassed the exitcode 0 to allow the Python code to continue running.


peterdragun (commented):
Overall LGTM, but we should also do the same thing for espefuse and espsecure.

hfudev:
Thank you. updated. PTAL again.

dobairoland:
@hfudev Thank you for this PR and looking into https://github.com/espressif/pytest-embedded/pull/362. Will it be enough for you if we create a dev release with this patch, i.e. 5.1.dev1?

dobairoland:
@peterdragun Could you please cherry-pick this into our internal repo?

hfudev:
> @hfudev Thank you for this PR and looking into [espressif/pytest-embedded#362](https://github.com/espressif/pytest-embedded/pull/362). Will it be enough for you if we create a dev release with this patch, i.e. 5.1.dev1?

Hi @roland can we release a 5.0.3 version? I think it's more like a bugfix instead of a feature.

dobairoland:
@hfudev Unfortunately, we already merged several features and the next release from the v5.* line will be 5.1.

dobairoland:
@hfudev This fix is available in https://pypi.org/project/esptool/5.1.dev1/. Thanks!

hfudev:
Thank you!

PR #1113, opened 2025-08-07
update Bluetooth string (ESPTOOL-1124)

shvmm:
change(esp32): Bluetooth description string from just `BT` to `BT (v4.2 BR/EDR) + BLE`

NO TESTING

NO TESTING


dobairoland:
I'm sorry but we won't accept this. First of all, this would be needed to specify for all chip targets and not just ESP32 only. Secondly, we cannot list every details in the log output - we could just list the entire datasheet there.

PR #1118, opened 2025-08-28
fix(esptool): Allow ELF files without segments again (ESPTOOL-1132)

tormodvolden:
The program header table can describe zero or more segments. However, commit ca16d5f3 introduced checks in _read_segments() that will fail on a zero number of program header entries.

ELF files created by objcopy based on a binary file may not have any segments / program header table. The number of program header entries e_phnum in the ELF header is zero.

This was fine before commit ca16d5f3, but is currently broken and esptool will exit with this error:
`A fatal error occurred: No segment header found at offset 0000 in ELF file.`

This PR reinstates the possibility of having no such entries by simply returning early and skip these checks in _read_segments() if e_phnum is zero.

The change has no implication if the number of entries is non-zero.

Tested with ELF files created by objcopy that have no segments. esptool elf2image now successfully creates an image.

NO_TESTING

### DIFF
--- a/esptool/bin_image.py
+++ b/esptool/bin_image.py
@@ -1434,6 +1434,8 @@ class ELFFile:
                 )
 
     def _read_segments(self, f, segment_header_offs, segment_header_count, shstrndx):
+        if segment_header_count == 0:
+            return
         f.seek(segment_header_offs)
         len_bytes = segment_header_count * self.LEN_SEG_HEADER
         segment_header = f.read(len_bytes)


Dzarda7:
Thanks @tormodvolden for the fix, will try to merge is ASAP.

Dzarda7:
@tormodvolden can you please change the commit message to something like `fix(elf2image): Handle ELF files with zero header counts`. Thanks.

tormodvolden:
Sure, done.

tormodvolden:
My original commit summary tried do convey that this fixes some kind of regression, since having zero program header counts was working fine in older versions of esptool. Actually "program header counts" would be better, what do you think?

Dzarda7:
> Actually "program header counts" would be better, what do you think?

 I am okay with that if you prefer. Will wait for your change. You can mention the regression in the body, but write `Allow ELF files without segments again` is not the best, especially when this worked like 4 years ago.

tormodvolden:
OK, I edited it to say "program headers".

Dzarda7:
@tormodvolden sorry, we just  merged some MR and we do not use merge commits, can you please rebase on master so you have credit in history?

tormodvolden:
Rebased to latest master.

PR #1121, opened 2025-09-15
rich-click 1.9 fix (ESPTOOL-1180)

dwreeves:
Resolves regression in rich-click 1.9.0 https://github.com/ewels/rich-click/issues/253

- rich-click 1.9.0 regression

NO TESTING

NO TESTING

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@
 </div>
 <hr>
 
+## v5.0.3 (2025-09-)
+
+- Fix bug with rich-click 1.9 compatibility *(Daniel Reeves - bd9241c)*
+
 ## v5.0.2 (2025-07-30)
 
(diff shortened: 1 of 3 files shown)


peterdragun (commented):
@dwreeves Thank you for the fix. This is much appreciated. Unfortunately, we cannot accept the PR, as we are following the conventional commits and have a bit more complicated release process (we use commitizen to automatically generate the Changelog based on the commits).

If you would like to fix the commit and take the credit, please update the commit message to something like: "fix(espefuse): Update CLI to support rich-click 1.9.0" and remove the changelog entry. 
If not I will take over, this is already quite valuable insight for us anyway.

dwreeves:
@peterdragun I'll close then and let you take it over, as I don't really care to take credit. I have a packed schedule today.

Thank you!

PR #1133, opened 2025-10-20
ci(github): Use new macos-15-intel runner for macOS builds

peterdragun:


peterdragun:
Sorry this was supposed to be in my fork 🙃

PR #1135, opened 2025-11-11
Fix typo in firmware image format documentation (ESPTOOL-1205)

tyeth:
Unchecked, but seems a sensible change.


tyeth:
There's also something weird in the next table, the three values are listed as '0xf', 0, 2 which might be another typo.
Feel free to close this and do a fresh commit with the correct title.

docs/en/advanced-topics/firmware-image-format.rst
```diff
@@ -94,7 +94,7 @@ The image header is 8 bytes long:
     +--------+------------------------------------------------------------------------------------------------+
     | 3      | High four bits - Flash size (``0`` = 1MB, ``1`` = 2MB, ``2`` = 4MB, ``3`` = 8MB, ``4`` = 16MB) |
     |        |                                                                                                |
-    |        | Low four bits - Flash frequency (``0`` = 80MHz, ``0`` = 40MHz, ``2`` = 20MHz)                  |
+    |        | Low four bits - Flash frequency (``0`` = 80MHz, ``1`` = 40MHz, ``2`` = 20MHz)                  |
```
peterdragun: I believe that this is correct and explained a couple of lines later, after the table in the note: `Flash frequency with value ``0`` can mean either 80MHz or 40MHz based on MSPI clock source mode.` Basically, there is a clock divider that will be set in ESP-IDF, which defines the final clock frequency.

dobairoland: I think this could be improved in a way so it won't raise suspicion about being a mistake. For example:

```
``0`` = 80MHz or 40MHz
```

PR #1139, opened 2025-11-24
fix(docs): Update zsh completion command syntax in documentation. (ESPTOOL-1229)

TinyuZhao:
NO TESTING, only involves document modification.


peterdragun:
Hi @TinyuZhao, thank you for creating the PR we already had a fix in review, which also included some other changes, so we decided to close this. Thanks for understanding.

PR #1140, opened 2025-11-24
fix(esp32c5): fix boot failure after USB-Serial-JTAG download (ESPTOOL-1230)

leeebo:
This commit resolves an issue where ESP32-C5 would fail to boot properly after firmware download via USB-Serial-JTAG interface, resulting in error `rst:0x7 (TG0_WDT_HPSYS),boot:0x18 (SPI_FAST_FLASH_BOOT).`

Using IDF command `idf.py flash monitor` download then restart `esp32-c5` through **USB-Serial-JTAG**, sometimes, the chip reboot in loop with

`"rst:0x7 (TG0_WDT_HPSYS),boot:0x18 (SPI_FAST_FLASH_BOOT)" ` 

The commit 1059ec7faff24ea1647c80963ef65abc515ef0fd try to fix this issue, But this fix did not fully work as intended.

I test with ESP32C5 ECO2 and ECO3, The `self.uses_usb_jtag_serial()` always return `176` when donwload through USB-Serial-JTAG, return `36` when using UART. which means the `self.uses_usb_jtag_serial()` is always `False` and the patch not work as normal

- [x] ESP32-C5 ECO2
- [x] ESP32-C5 ECO3
- [x] esptool v5.1.0

**I believe this fix needs to be backported to other branches that support esp32-c5.**

### DIFF
--- a/esptool/targets/esp32c5.py
+++ b/esptool/targets/esp32c5.py
@@ -60,7 +60,7 @@ class ESP32C5ROM(ESP32C6ROM):
     PCR_SYSCLK_XTAL_FREQ_V = 0x7F << 24
     PCR_SYSCLK_XTAL_FREQ_S = 24
 
-    UARTDEV_BUF_NO = 0x4085f514  # Variable in ROM .bss which indicates the port in use
+    UARTDEV_BUF_NO = 0x4085F514  # Variable in ROM .bss which indicates the port in use
 
     FLASH_FREQUENCY = {
         "80m": 0xF,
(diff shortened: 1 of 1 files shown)


esptool/targets/esp32c5.py
```diff
@@ -61,6 +61,7 @@ class ESP32C5ROM(ESP32C6ROM):
     PCR_SYSCLK_XTAL_FREQ_S = 24
 
     UARTDEV_BUF_NO = 0x4085F51C  # Variable in ROM .bss which indicates the port in use
```
leeebo: I realy not sure if the address of `UARTDEV_BUF_NO ` is correct

radimkarnis:
Hello @leeebo,
thanks for opening this PR. I can reproduce the issue, it really seems like the `UARTDEV_BUF_NO` offset is wrong. `UARTDEV_BUF_NO_USB_JTAG_SERIAL` has to be `3`, not `176`.

I only have ECO2 with functional USB-Serial/JTAG mode. Testing with that, changing the `UARTDEV_BUF_NO` to `0x4085f514` solves the issue. Could you please help me verify this with the ECO3? Thank you

leeebo:
Hi @radimkarnis , `changing the UARTDEV_BUF_NO to 0x4085f514` works as expected on ESP32-C5 ECO3.

Could you please help us verify the UARTDEV_BUF_NO address of `ESP32-C61`? Because we have reproduced the same issue on ESP32-C61.

By the way, regarding the BSS address — will it remain unchanged after a ROM ECO? If not, do we need to locate the address based on the specific ECO version of the chip?

radimkarnis:
@leeebo 

> Could you please help us verify the UARTDEV_BUF_NO address of ESP32-C61

Please try and verify `0x4084f5ec` for ECO2 and `0x4084f5e4` for ECO3.

> By the way, regarding the BSS address — will it remain unchanged after a ROM ECO? If not, do we need to locate the address based on the specific ECO version of the chip?

No, as you can see, we cannot guarantee the address won't change with new ECO versions. 

If you verify the C61, I can do a quick fix and release a dev version that would propagate into IDF.

PR #1157, opened 2026-04-07
Add workaround to support CP2102C (ESPTOOL-1303)

Jeija:
The [CP2102C](https://www.silabs.com/interface/usb-bridges/driverless-usb-bridges) is a new USB-to-UART bridge chip by Silicon Labs. It is not to be confused with the CP2102N, which is an older and completely different chip.

The CP2102C is quite attractive due to its low price (only around 1€ at sufficiently large quantities, vs 2-3€ for chips like the CP2102N), and it is supported by the CDC ACM driver, but it has one major issue: It has hardware flow control enabled by default (see [errata sheet](https://www.silabs.com/documents/public/errata/cp2102c-errata.pdf)), and this cannot be disabled.

In practice, when using the CP2102C with an ESP32, this means two things:
* The CTS pin must be tied to GND so that the CP2102C transmits anything at all. This is easy to ensure in hardware.
* From what I can tell, the RTS pin is pulled low whenever the CP210C receives something on its RX port and DTR is high. From my understanding there is nothing the computer can do about it. This completely breaks the reset sequence required for the ESP32, but this pull requests adds a workaround for this in software.

### Problem description
This is what happens without this workaround:
Screenshot From 2026-04-07 14-41-27

BOOT is low, CHIP_PU=EN=~RST goes high, ESP32 enters boot mode. The problem: Whenever the ESP32 sends something while DTR is high, RTS goes low, which means CHIP_PU goes low, so the ESP32 resets and we can never successfully flash anything.

With this workaround, we get this instead (as it should be):
Screenshot From 2026-04-07 14-43-08

The only change: We do not set DTR=HIGH at the end of the reset sequence, but just keep it low.

Screenshot From 2026-04-07 14-22-11
With the "standard" reset circuitry like in the picture above, this ensures that we can flash the firmware correctly.

I understand that this is quite an "ugly" fix and that something similar can also be achieved simply using a custom reset sequence. It might nevertheless make sense to include a fix like this in esptool, as more and more people might start running into this issue now that the CP2102C is getting more common.

### I have tested this change with the following hardware & software combinations:
ESP32-D0WD-V3, CP2102C

### I have run the esptool automated integration tests with this change and the above hardware:
NO TESTING

### DIFF
--- a/esptool/loader.py
+++ b/esptool/loader.py
@@ -373,6 +373,9 @@ class ESPLoader:
     ESPRESSIF_VID = 0x303A
 
     USB_JTAG_SERIAL_PID = 0x1001
+    HARDWARE_FLOW_CONTROL_VID_PIDS = [
+        (0x10C4, 0xEA64),  # SiLabs CP2102C USB to UART Bridge Controller
+    ]
 
     # Chip IDs that are no longer supported by esptool
     UNSUPPORTED_CHIPS = {
@@ -812,18 +815,20 @@ class ESPLoader:
         if mode == "usb-reset" or self.get_usb_vid_pid()[1] == self.USB_JTAG_SERIAL_PID:
             return (USBJTAGSerialReset(self._port),)
 
+        flow_control = self.uses_hardware_flow_control()
+
         # USB-to-Serial bridge
         if os.name != "nt" and not self._port.name.startswith("rfc2217:"):
             return (
-                UnixTightReset(self._port, delay),
-                UnixTightReset(self._port, extra_delay),
-                ClassicReset(self._port, delay),
-                ClassicReset(self._port, extra_delay),
+                UnixTightReset(self._port, delay, flow_control),
+                UnixTightReset(self._port, extra_delay, flow_control),
+                ClassicReset(self._port, delay, flow_control),
+                ClassicReset(self._port, extra_delay, flow_control),
             )
 
         return (
-            ClassicReset(self._port, delay),
-            ClassicReset(self._port, extra_delay),
+            ClassicReset(self._port, delay, flow_control),
+            ClassicReset(self._port, extra_delay, flow_control),
         )
 
     def connect(
@@ -1241,6 +1246,14 @@ class ESPLoader:
         True if the host sees this port as Espressif USB-OTG (VID/PID match).
         """
         return self.get_usb_vid_pid() == (self.ESPRESSIF_VID, self.IMAGE_CHIP_ID)
+
+    def uses_hardware_flow_control(self):
+        """
+        True if this port uses hardware flow control that esptool needs to know
+        about for resetting. This is determined by checking if the USB PID matches
+        known PIDs with hardware flow control.
+        """
+        return self.get_usb_vid_pid() in self.HARDWARE_FLOW_CONTROL_VID_PIDS
 
     def get_usb_mode(self):
         """
@@ -2044,7 +2057,9 @@ class ESPLoader:
         if cfg_custom_hard_reset_sequence is not None:
             CustomReset(self._port, cfg_custom_hard_reset_sequence)()
         else:
-            HardReset(self._port, uses_usb)()
+            HardReset(
+                self._port, uses_usb, flow_control=self.uses_hardware_flow_control()
+            )()
 
     def soft_reset(self, stay_in_bootloader):
         if not self.IS_STUB:
--- a/esptool/reset.py
+++ b/esptool/reset.py
@@ -30,9 +30,10 @@ DEFAULT_RESET_DELAY = 0.05  # default time to wait before releasing boot pin aft
 class ResetStrategy:
     print_once = PrintOnce(log.warning)
 
-    def __init__(self, port, reset_delay=DEFAULT_RESET_DELAY):
+    def __init__(self, port, reset_delay=DEFAULT_RESET_DELAY, flow_control=False):
         self.port = port
         self.reset_delay = reset_delay
+        self.flow_control = flow_control
 
     def __call__(self):
         """
@@ -89,6 +90,17 @@ class ResetStrategy:
             status &= ~TIOCM_RTS
         fcntl.ioctl(self.port.fileno(), TIOCMSET, struct.pack("I", status))
 
+    def _setHUPCL(self, enabled):
+        if os.name == "nt" or not hasattr(termios, "HUPCL"):
+            return False
+        attrs = termios.tcgetattr(self.port.fileno())
+        if enabled:
+            attrs[2] |= termios.HUPCL
+        else:
+            attrs[2] &= ~termios.HUPCL
+        termios.tcsetattr(self.port.fileno(), termios.TCSANOW, attrs)
+        return True
+
 
 class ClassicReset(ResetStrategy):
     """
@@ -102,7 +114,11 @@ class ClassicReset(ResetStrategy):
         self._setDTR(True)  # IO0=LOW
         self._setRTS(False)  # EN=HIGH, chip out of reset
         time.sleep(self.reset_delay)
-        self._setDTR(False)  # IO0=HIGH, done
+        if not self.flow_control:
+            # We usually want to put IO0 back high here, but we can't do that with
+            # some hardware-flow control enabled adapters (like the CP2102C),
+            # as they will otherwise pull RTS low when receiving data from the ESP32.
+            self._setDTR(False)  # IO0=HIGH, done
 
 
 class UnixTightReset(ResetStrategy):
@@ -118,8 +134,12 @@ class UnixTightReset(ResetStrategy):
         time.sleep(0.1)
         self._setDTRandRTS(True, False)  # IO0=LOW & EN=HIGH, chip out of reset
         time.sleep(self.reset_delay)
-        self._setDTRandRTS(False, False)  # IO0=HIGH, done
-        self._setDTR(False)  # Needed in some environments to ensure IO0=HIGH
+        if not self.flow_control:
+            # We usually want to put IO0 back high here, but we can't do that with
+            # some hardware-flow control enabled adapters (like the CP2102C),
+            # as they will otherwise pull RTS low when receiving data from the ESP32.
+            self._setDTRandRTS(False, False)  # IO0=HIGH, done
+            self._setDTR(False)  # Needed in some environments to ensure IO0=HIGH
 
 
 class USBJTAGSerialReset(ResetStrategy):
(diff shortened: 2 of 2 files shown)


radimkarnis:
Hi @Jeija,
thanks for the investigation and the nice PR! We'll have an internal discussion about this.

a preliminary note:
- I believe it would make more sense not to add a new CP2102C-specific reset strategy, but to extend the `ClassicReset` strategy instead not to run the last `self._setDTR(False)` call when `self.flow_control` is `True`.
    - `self.flow_control` would be defined in the `ResetStrategy` parent class and would be set based on `CP2102C_PID` detection, the way you do it now
    - this would allow us to extend the fix to other reset sequences if needed (is not `UnixTightReset` affected as well?)

Jeija:
@radimkarnis Thanks for the feedback, I definitely agree!

I just added new commits that implement this in a more generic way, as you suggested (tested both `UnixTightReset` and `ClassicReset`). I also fixed `HardReset` on the CP2102C - unfortunately, the only way I found to make it work is to wait >1s in the end before closing the port, which is a pretty ugly workaround, but I haven't found any alternative so far.

esptool/loader.py
```diff
@@ -783,18 +786,22 @@ def _construct_reset_strategy_sequence(self, mode):
...
+        flow_control = self._get_pid() in self.HARDWARE_FLOW_CONTROL_PIDS
+
         # USB-to-Serial bridge
         if os.name != "nt" and not self._port.name.startswith("rfc2217:"):
             return (
-                UnixTightReset(self._port, delay),
-                UnixTightReset(self._port, extra_delay),
-                ClassicReset(self._port, delay),
-                ClassicReset(self._port, extra_delay),
+                UnixTightReset(self._port, delay, flow_control=flow_control),
```
radimkarnis, line 796: Please pass the option simply as `flow_control` instead of `flow_control=flow_control` to keep things consistent.

esptool/reset.py
```diff
@@ -102,7 +104,10 @@ def reset(self):
         self._setDTR(True)  # IO0=LOW
         self._setRTS(False)  # EN=HIGH, chip out of reset
         time.sleep(self.reset_delay)
-        self._setDTR(False)  # IO0=HIGH, done
+        if not self.flow_control:
+            # Flow control: Keep DTR high so that RTS toggling doesn't cause
```
radimkarnis, line 108: nitpick: the new comment says “Flow control: Keep DTR high …” but it sits inside if not self.flow_control:, which is the branch for non flow-control adapters.

esptool/reset.py
```diff
@@ -149,21 +157,33 @@ class HardReset(ResetStrategy):
...
+                time.sleep(0.2)
+            else:
+                time.sleep(0.1)
+                self._setRTS(False)
         else:
+            self._setDTR(False)
+            self._setRTS(True)
             time.sleep(0.1)
-            self._setRTS(False)
+            self._setRTS(True)
```
radimkarnis, line 182: There are now two `self._setRTS(True)` calls in a row (with a 0.1 delay between them)

esptool/reset.py
```diff
@@ -149,21 +157,33 @@ class HardReset(ResetStrategy):
...
+            if self.uses_usb:
+                # Give the chip some time to come out of reset,
+                # to be able to handle further DTR/RTS transitions
+                time.sleep(0.2)
+                self._setRTS(False)
+                time.sleep(0.2)
+            else:
+                time.sleep(0.1)
+                self._setRTS(False)
         else:
```
radimkarnis, line 178: I believe this case requires some more attention or explanations.
Why is `FLOW_CONTROL_HOLD_TIME` so unreasonably high? Why is this case needed at all?

esptool/reset.py
```diff
@@ -149,21 +157,33 @@ class HardReset(ResetStrategy):
...
+            self._setDTR(False)
+            self._setRTS(True)
             time.sleep(0.1)
-            self._setRTS(False)
+            self._setRTS(True)
+            self._setDTR(True)
+            # Give the target time to emit enough data so RTS doesn't toggle
+            # back during the most fragile part of boot.
+            time.sleep(FLOW_CONTROL_HOLD_TIME)
```
radimkarnis, line 186: The normal HardReset path ends by asserting RTS low again (_setRTS(False) at lines 173 and 177). The flow_control path never calls _setRTS(False) - it ends after FLOW_CONTROL_HOLD_TIME with RTS still driven True in software terms. That may be acceptable if esptool always closes the port immediately after (so the bridge returns lines to a benign idle), but it differs from the classic contract and could surprise anyone calling hard_reset() on a long-lived ESPLoader instance.

radimkarnis (commented):
Thanks for accepting my suggestion! We will go forward with this PR, but I am not a fan of the long wait at the end of Hard Reset.

esptool/loader.py
Jeija: Agreed, changed :+1:

esptool/reset.py
Jeija: Yes, the wording here was quite confusing (though technically correct), I tried to rephrase this in the latest version.

esptool/reset.py
Jeija: This `FLOW_CONTROL_HOLD_TIME` was a really ugly workaround that I didn't find a better alternative for at the time. After some digging, I have now found that setting HUPCL on Linux also works, and that the wait time was not necessary on Windows after all. Seems to have something to do with what Linux does when closing the port as far as I can tell.

esptool/reset.py
Jeija: I updated the code to fix this, agreed.

Jeija:
Thanks for your feedback and especially the pushback regarding the ugly long 1.5s wait during hard reset. That challenged me to find a better solution, which seems to be to write to the `HUPCL` on Unix. I have tried to address your comments in my recent couple of commits.

radimkarnis:
@Jeija thanks for the nice progress! 
- Please see the Copilot review
- the `_get_pid()` API has unfortunately changed in the meantime - got replaced with `get_usb_vid_pid()`

esptool/loader.py
```diff
@@ -783,18 +786,22 @@ def _construct_reset_strategy_sequence(self, mode):
         if mode == "usb-reset" or self._get_pid() == self.USB_JTAG_SERIAL_PID:
             return (USBJTAGSerialReset(self._port),)
 
+        # Some chips have hardware flow control that cannot be disabled,
+        # and this can interfere with reset if esptool doesn't know about it.
+        flow_control = self._get_pid() in self.HARDWARE_FLOW_CONTROL_PIDS
+
```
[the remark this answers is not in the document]
Jeija: This should be fixed by using the new `get_usb_vid_pid` function, which has a cache.

esptool/reset.py
```diff
@@ -149,21 +169,41 @@ class HardReset(ResetStrategy):
...
+            self._setRTS(True)  # EN->LOW
+            time.sleep(0.1)
+            self._setDTR(True)
             time.sleep(0.1)
             self._setRTS(False)
+            if not has_hupcl:
+                # Windows has no HUPCL equivalent. Release RTS first while
+                # DTR is still asserted, then release DTR to leave the reset
+                # circuit idle before close.
+                self._setDTR(False)
```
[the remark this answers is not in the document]
Jeija: For me, when connecting the CP2102C to my computer, HUPCL is set to true by default (`stty -F /dev/ttyACM0 -a` shows `hupcl`, not `-hupcl`). In order for the HardReset strategy to work without the long delay, it must be set to false when closing. It is insufficient to only set it to false during the close and then restore the value afterwards: HUPCL must remain false also after closing the port. Otherwise, RTS is immediately pulled low when setting HUPCL true (restoring the previous value) and the device goes into reset mode. So fundamentally, this is impossible to solve without the delay and without side effects on other software: The serial port will have HUPCL set to false afterwards.

One could argue that `esptool.py` should not have side effects and that users must manually execute `stty -F /dev/ttyACM0 -hupcl` before using hardware flow control-enabled adapters, but I think just setting HUPCL to false is the lesser evil.

esptool/reset.py
```diff
@@ -149,21 +169,41 @@ class HardReset(ResetStrategy):
...
+            self._setDTR(False)  # IO0=HIGH
+            self._setRTS(True)  # EN->LOW
+            time.sleep(0.1)
+            self._setDTR(True)
             time.sleep(0.1)
             self._setRTS(False)
+            if not has_hupcl:
+                # Windows has no HUPCL equivalent. Release RTS first while
+                # DTR is still asserted, then release DTR to leave the reset
+                # circuit idle before close.
```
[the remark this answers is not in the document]
Jeija: The suggested hard reset strategy does not work with real hardware. I think copilot does not understand the reset circuitry context.

Jeija:
I updated my PR to use the new `get_usb_vid_pid()` function, which also solves the VID/PID caching issue (and is much nicer, because VID/PID is obviously more unique than just PID).
Regarding the other suggestions by Copilot, they unfortunately do not work in practice, see my comments above.

radimkarnis:
Hi @Jeija, awesome work, thanks again!

If you squash the commits and fix the pre-commit failing check (some lines are longer than limit), we can proceed to put this through our internal review queue and hopefully merge very soon.

radimkarnis:
@Jeija I understand I asked for many revisions in this PR, let me know if you are not interested in following up with this. I can finish it. Thanks!

Jeija:
@radimkarnis No worries, I was planning to finish it this week. Just didn't find the time last week and today I don't have all the hardware here for testing. The remaining changes (squashing and making the linter happy) are easy enough, just wanted to test the final version with real hardware again to be safe.

Jeija:
@radimkarnis I just squashed the commits into one on this branch (for reference, see https://github.com/Jeija/esptool/tree/cp2102c-backup for the full commit history), fixed the line length issue and re-tested the change with both a CP2102C and a non-hardware flow control adapter (FT231X). Should be good now.

radimkarnis:
@Jeija Thank you, nice work! 
I have now put this through our internal testing and review queue. Hopefully, this will get merged in a breeze.

esptool/reset.py
```diff
@@ -149,21 +169,41 @@ class HardReset(ResetStrategy):
...
+            # Keep DTR asserted across close so RTS flow-control changes can't reset
+            # the chip during boot chatter.
+            has_hupcl = self._setHUPCL(False)
+            self._setDTR(False)  # IO0=HIGH
+            self._setRTS(True)  # EN->LOW
+            time.sleep(0.1)
+            self._setDTR(True)
             time.sleep(0.1)
             self._setRTS(False)
+            if not has_hupcl:
```
radimkarnis, line 189: would be nice to explain why the HUPCL makes the difference here (at least for me it is not obvious)

esptool/reset.py
```diff
@@ -149,21 +169,41 @@ class HardReset(ResetStrategy):
     Can be used to reset out of the bootloader or to restart a running app.
     """
 
-    def __init__(self, port, uses_usb=False):
-        super().__init__(port)
+    def __init__(
+        self, port, uses_usb=False, reset_delay=DEFAULT_RESET_DELAY, flow_control=False
```
radimkarnis: nitpick: `reset_delay` was added here but it is not used anywhere

radimkarnis (commented):
I have additional questions or nitpicks from my colleagues:

IIUC, this might cause issues with external reset, right? When the chip is sending something over UART (which most of the time will), and an external reset is triggered (using a button, for example), the chip will enter download mode, right?

Jeija:
> IIUC, this might cause issues with external reset, right? When the chip is sending something over UART (which most of the time will), and an external reset is triggered (using a button, for example), the chip will enter download mode, right?

Yes, this is correct, but that seems to be a fundamental issue with hardware flow control adapters from what I can tell. This also happens if i just connect the ESP32 via the CP2102C to a computer and don't execute any command (no esptool, no serial port terminal). The reset button always resets the ESP32 into download boot mode, proper reset is only possible via a reset sequence or by power cycling the device.

esptool/reset.py
Jeija: Good point, I removed this in my latest change to the squashed commit.

esptool/reset.py
Jeija: Here are some empirical results (measurements of DTR / RTS / BOOT=IO0 / ~RESET), measured with a Saleae Logic Pro and for command `./esptool.py -p /dev/ttyACM0 -a hard-reset chip-id`.

with modified HUPCL

This is what I observed with modified HUPCL (reset works fine).

umodified HUPCL

This is with unmodified HUPCL. Notice how closing the port after the hard reset procedure ends with RTS low and DTR high, which causes RESET to go low and thus the ESP32 to enter a permanent RESET state. From my understanding, HUPCL determines what the USB-UART chip is supposed to do when the connection / port is closed.

radimkarnis:
@Jeija, although there are certain caveats to the approach, let's try and see how it goes and what future USB-to-UART bridges require.

Thank you for the contribution and your patience!

Issue #1015, opened 2024-09-27
Clarify supported SPI flash for use with --spi-connection (ESPTOOL-931)


nebkat:
### Is your feature request related to a problem?

I have been trying to no avail to get a variety of NAND flash chips to work with `esptool read_flash`. My understanding is that NAND flash is not supported in ROM due to its complexity?

The [Overriding SPI Flash Connections](https://docs.espressif.com/projects/esptool/en/latest/esp32/esptool/advanced-options.html#overriding-spi-flash-connections) documentation does not sufficiently explain the requirements for interfacing with an SPI flash chip.

### Describe the solution you'd like

Could we get clear documentation on what type of chips are supported, as many are advertised as "SPI flash".

### Describe alternatives you've considered

Could a custom flasher stub be made that is able to communicate with more types of flash?

radimkarnis:
Hello @nebkat,

NAND flash chips are not supported, the ROM doesn't implement the needed driver. You need a NOR flash chip capable of at least DIO mode.

We don't keep a list of all supported chips. You can see the [flashrom source code](https://github.com/flashrom/flashrom/blob/master/include/flashchips.h) to see the most common flash chips. Other extra modes, such as OPI flash and 32-bit addressing are described [here](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/peripherals/spi_flash/spi_flash_optional_feature.html#optional-features-for-flash), but that doesn't include NAND flash chips.

I will add a note to esptool docs to explain this.

> Could a custom flasher stub be made that is able to communicate with more types of flash?

It could, but that would only help with flashing and reading the data with esptool. If you wanted to use the flash memory in your app, you would have to re-implement the driver in your code, because the ROM, ESP-IDF and Arduino don't support it. There might also be other reasons why it's not supported I am not aware of (e.g. HW limitations).

nebkat:
Hey @radimkarnis,

Thank you for confirming my suspicions - yes if a note could be added I hope it will save someone else some confusion!

> It could, but that would only help with flashing and reading the data with esptool.

This is what I would be most interested in. We use NAND flash in all of our products for user data storage, would be amazing if we could pull data using an esptool-like interface. Perhaps something based on https://github.com/espressif/idf-extra-components/tree/master/spi_nand_flash and `esptool load_ram`. I will look in to it!

PR #1166, opened 2026-05-31
feat(stub_flasher): Add SDMMC card support on ESP32-S3 (ESPTOOL-1330)

nebkat:
# **_Proof of concept_*

Adds host-side support for the new "sdmmc" plugin opcodes (0xDF–0xE5) shipped by the esp-flasher-stub.  Mirrors the NAND plugin integration: plugin-aware CLI, ESPLoader wrapper methods, and dispatch in write_flash / read_flash / erase_flash / erase_region / attach_flash.

```
CLI:
  esptool sdmmc-info                                 # smoke test attach
  esptool --flash-type sdmmc write-flash 0 image.bin # sector-aligned writes
  esptool --flash-type sdmmc read-flash 0 0x10000 out.bin
  esptool --flash-type sdmmc erase-region 0 0x10000
```

Optional overrides via `--sdmmc-slot` / `--sdmmc-width` / `--sdmmc-freq-khz` / `--sdmmc-pins`; with all defaults the stub uses IDF v6.0 SDMMC_SLOT_CONFIG_DEFAULT for ESP32-S3 (CLK=14, CMD=15, D0=2, D1=4, D2=12, D3=13, D4=33, D5=34, D6=35, D7=36, slot 0, 4-bit, 20 MHz).

On error `sdmmc_attach` decodes a packed {stage, err, RINTSTS} word from the stub's response so failures surface as e.g. "SDMMC attach failed at CMD8: SDMMC_ERR_TIMEOUT; RINTSTS=0x0100 [RTO]; stub status C400" instead of a generic SPI op failure.

Also bumps the bundled `stub_flasher/2/esp32s3.json` to the matching `esp-flasher-stub` `feat/sdmmc` build.

Full disclosure: quick and dirty AI attempt, not heavily tested but appears to be working. ESP32S3 custom board with SDMMC NAND.

NO TESTING


dobairoland:
Hi @nebkat. Thank you for the contribution. I see that this isn't ready, but here are some heads-up.

The implementation itself is clean and consistent — it reuses the existing plugin architecture, so integration risk is low. My hesitation is more fundamental: whether SDMMC belongs in esptool at all. Even once this is polished (CI, tests, HW validation), merging hinges on the use case, so I'd like to understand that first.

**Could you describe the concrete scenario?** Is the card removable or soldered (eMMC/SD-NAND)? What image sizes and write times are you working with? The reason I ask is that the existing approaches likely cover most situations better:

- **Removable card** — pulling it and using a PC card reader is faster and needs no ESP at all.
- **Bulk data onto a soldered card** — a small USB-MSC loader firmware (e.g. the ESP-IDF `tinyusb` MSC example) exposes the card as a block device at full card speed. This PR instead funnels every 512-byte block through the UART/USB-*serial* protocol (single-block, polled, window of 1, 20 MHz bus), so a multi-GB card takes hours rather than minutes — the wrong tool for bulk writes.
- **Conceptually**, esptool flashes the chip's own boot media; an SD card is application storage, so `write-flash --flash-type sdmmc` stretches the tool's scope.

The one niche I *can* see is provisioning a **small** image (partition table, small FAT/SPIFFS, config blobs) onto a soldered card over the same cable already used for SPI flash — "one tool, one step." If that's your case, it'd help to say so with the image sizes involved.

Two more questions: is S3-only the intended scope (ESP32-P4 has the stronger SDMMC story)? And would documenting the USB-MSC approach meet your need instead?

Not a no — just that here the justification matters more than the code, and I'd rather settle that before you turn a PoC into production-quality work.

---
_Generated by [Claude Code](https://claude.ai/code)_

nebkat:
Hi @dobairoland,

We have many devices in the wild with soldered 512MB eMMC flash holding user data.

Occasionally devices get returned due to flash failure or FAT corruption for us to recover. Thus far we have been flashing a custom firmware which exposes the raw contents as a dump file over WiFi hotspot but it involved overwriting the existing firmware and was not very user friendly.

Other times we are refurbishing devices and need do perform a quick format of the flash before sending to the next customer - for which we have yet to develop a good solution.

Our new firmware uses a partition table and multiple proprietary sequential block based logging systems on the SDMMC so we get a lot of useful information out from reading just a handful of blocks at fixed offsets.

----

I have long been [interested](https://github.com/espressif/esptool/issues/1015) in this ability as it greatly simplifies the above. The benefits I see are:
- Re-use existing esptool infrastructure loading/flashing/commands/plugins/etc
- Intuitive for anyone familiar with esptool
- Fits in neatly with existing esptool based provisioning pipeline
- Slow but reliable (important in our use case when dealing with critical user data)
- Simple enough that our more technical dealers could even perform a dump or format themselves with just `pip install esptool`, no proprietary tools

To me this is a natural extension of the flashing capabilities of `esptool`.

S3 just happens to be what we are using and had the existing NAND plugin. We also have some older ESP32 devices and we are waiting for you guys to drop an FCC certified ESP32-P4 + C6 module :) 

----

All that said your USB-MSC idea is very interesting and I am exploring this now, thank you for the suggestion! I think you are right that this will be the correct approach for bulk reading, with the plugin remaining useful more for single connection provisioning of small images.

----

Aside: it would be awesome if even more peripherals such as GPIO, UART or I2C could be flexed through the new plugin system. This would simply be _invaluable_ for testing boards and provisioning other peripherals (e.g. GPS receivers). While it may be outside the current scope of esptool, it would benefit hugely from the existing infrastructure it provides..

In fact this would open up a whole new selling point for ESP32s - quick and dirty GPIO/UART control of any ESP device!

radimkarnis:
@nebkat thanks for the PoC and explanation!

I see why this can be very useful for your case. 
Unfortunately, I share Roland's views and consider it to be very niche for us to accept and maintain the additional 1500 lines of code across multiple repositories. 

> peripherals such as GPIO, UART or I2C could be flexed through the new plugin system

Using esptool as a CLI and an interface to send commands from the host and the flasher stub as a low-level means to expose all of the chip registers and peripherals is an infinitely scalable concept. But I believe that's where the line has to be drawn. We provide the means and documentation for the users to build their own plugins and expand the functionality for their use case. 

Our goal is to allow the users to do that (that's why the plugin system and overall flasher stub development are documented), while maintaining the core esptool functions and adding/debugging the latest support for Espressif chips. 

If the feature is heavily requested, we can consider it to become a part of the core set. Currently, we don't know about any other similar feature requests.

nebkat:
Understood, thank you for the feedback! That is a reasonable compromise, I will continue developing this as a set of third party tools.

nebkat:
If you could consider a small request: would be great if plugins consumed a single opcode and then had a 4-byte subcommand in their payload (i.e. NAND = 0xD5, attach = 0x00000001). I'd like to maintain support for the built in plugins going forward but I see the opcode space is rather tight.

radimkarnis:
@nebkat we'll consider the request and are tracking it internally now. Thanks.

PR #1171, opened 2026-06-17
fix: resolve symlinked paths in get_usb_vid_pid (ESPTOOL-1342)

horw:
Added a fallback to try resolving the symlink PID and VID.

### DIFF
--- a/esptool/loader.py
+++ b/esptool/loader.py
@@ -705,6 +705,8 @@ class ESPLoader:
                 "COM and /dev/ serial ports."
             )
             return None, None
+
+        symlink_port = active_port
         # Return the real path if the active port is a symlink
         if active_port.startswith("/dev/") and os.path.islink(active_port):
             active_port = os.path.realpath(active_port)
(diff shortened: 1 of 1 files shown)


Dzarda7:
Thanks @horw, this seems as nice fix to me. I have no issue with that, only thing, codewise I believe it can be unified with the block above.

But we plan to use [esp-pylib](https://github.com/espressif/esp-pylib) for these common things we use across multiple tools, so we might close this and put it directly there. My question is @peterdragun, do you see integration of pylib into esptool happening soon (in the matter of days) so we can do it the way I propose or rather merge it in esptool?

peterdragun:
Correct - I will add this to esp-pylib as well. IMO, it is not worth merging here; the esp-pylib migration is already in process, and hopefully, in a matter of days, we will be able to review and merge it. 

Might be worth considering adding this to the v4 branch, though.

ydhub:
just FYI,  May use `include_links=True` for `comports` can simplify this method? 
... 
or maybe this MR is better. `include_links` only process links inside `/dev`

radimkarnis (approved):
Nice addition, thanks for the PR!

Dzarda7:
Merged to [esp-pylib](https://github.com/espressif/esp-pylib/pull/36), should be in master soon, merged into v4 branch in https://github.com/espressif/esptool/commit/80684fd0456ed8083ca6856eaf7133e844e5d3cd.

PR #1185, opened 2026-09-13
feat: Add --no-reset option to esp_rfc2217_server (ESPTOOL-1382)

Josverl:
In testing [MicroPython's mpremote](https://pypi.org/project/mpremote/) to connect to an ESP32-S3 devboard using the `esp_rfc2217_server` no REPL connection can be established as the board will be reset into bootloader mode on connection.
mpremote connects successfully to the same board over a serial or USB CDC connection without triggering the bootloader.

This pr adds a `--no-reset` flag to the tool that prevents the automatic reset.
The default for the flag is off to maintain backward compatibility.

- `python -m esp_rfc2217_server COM71 -p 22170 --no-reset`
- `mpremote connect rfc2217://localhost:22170`

<details><summary>--no-reset behavior</summary>
<p>

(micropython) PS C:\dev\micropython\scratch\esptool> python -m esp_rfc2217_server COM71 -p 22170 --no-reset -vvv
RFC 2217 TCP/IP to Serial redirector - type Ctrl-C / BREAK to quit
Serving serial port: COM71
TCP/IP port: 22170
Waiting for connection ... use the 'rfc2217://192.168.1.18:22170?ign_set_control' as a PORT
.Connected by 127.0.0.1:61781
INFO: client accepts RFC 2217
INFO: NOTIFY_MODEMSTATE: False
INFO: client accepts RFC 2217
INFO: NOTIFY_MODEMSTATE: False
INFO: set baud rate: 115200
INFO: set data size: 8
INFO: set parity: N
INFO: set stop bits: 1
INFO: changed flow control to None
INFO: purge in
INFO: purge out

</p>
</details> 

<details><summary>Old behavior</summary>
<p>

python -m esp_rfc2217_server COM71 -p 22170 -vvv             
RFC 2217 TCP/IP to Serial redirector - type Ctrl-C / BREAK to quit
Serving serial port: COM71
TCP/IP port: 22170
Waiting for connection ... use the 'rfc2217://192.168.1.18:22170?ign_set_control' as a PORT
..Connected by 127.0.0.1:51993
INFO: client accepts RFC 2217
INFO: NOTIFY_MODEMSTATE: False
INFO: client accepts RFC 2217
INFO: NOTIFY_MODEMSTATE: False
INFO: set baud rate: 115200
INFO: set data size: 8
INFO: set parity: N
INFO: set stop bits: 1
INFO: changed flow control to None
INFO: Activating reset in thread

</p>
</details> 

### I have tested this change with the following hardware & software combinations:
Tested on Windows 11:
ESP32, ESP32-S3 , ESP32-C3

### I have NOT run the esptool automated integration tests with this change and the above hardware:

LIMITED TESTING has been performed , as the test suite throws many errors on Windows that appear unrelated to the this change.


dobairoland:
Thank you for the PR. If I understood your problem well, then you are trying to solve the chip being put into download mode all the time (by all clients, not just esptool, which really needs it). Your PR is trying to fix the symptoms, but I think the real solution would be to address the issue in the server. I've put together a PoC which aims to address this.

Most dev boards connect the USB-UART bridge to the chip through this auto-reset circuit (1 = line asserted):

| DTR | RTS | EN | GPIO0 | Chip |
|:-:|:-:|:-:|:-:|---|
| 1 | 1 | high | high | runs |
| 0 | 0 | high | high | runs |
| 0 | 1 | **low** | high | held in reset |
| 1 | 0 | high | **low** | enters download mode if EN was low just before |

The current server starts the bootloader reset whenever a client sends DTR=1. pyserial's RFC 2217 client sends DTR=1 and RTS=1 every time it opens the port, so every client resets the chip into download mode. On a local serial port the same line states keep the chip running.

The PoC does not forward DTR/RTS requests to the serial port. It stores them and tracks what EN would do on the circuit above. A reset sequence starts only after the client holds EN low and then releases it. Any other request only updates the stored state.

```mermaid
stateDiagram-v2
    state "EN high" as released
    state "EN low" as held
    [*] --> released: client connects
    released --> held: RTS=1 and DTR=0
    held --> released: DTR=1 → bootloader reset
    held --> released: RTS=0 → hard reset
```

How common clients are handled:

| Client action | DTR/RTS requests | Current server | PoC |
|---|---|---|---|
| Opens the port with default line states (serial terminal, mpremote) | DTR=1, RTS=1 | bootloader reset | no reset |
| Opens the port with both lines released (esptool on Windows) | DTR=0, RTS=0 | hard reset | no reset |
| esptool bootloader reset | DTR=0, RTS=1, then DTR=1, RTS=0 | bootloader reset | bootloader reset |
| esptool hard reset | RTS=1, then RTS=0 while DTR=0 | hard reset | hard reset |

The server now also answers the DTR/RTS requests, so `?ign_set_control` is no longer needed in the port URL.

But this needs more careful investigation.

Josverl:
@dobairoland 
Note the there is [specific management of DTR/RTS in mpremote on Windows](https://github.com/micropython/micropython/blob/52b5fbcb4b7961356c83ba0c533ee42acdfa0169/tools/mpremote/mpremote/transport_serial.py#L72-L85) compared to manage the different OS handling of these signals without triggering unneeded hard/soft resets on connection and/or disconnection.
Therefore your state table may need at least one additional dimension:  the OS and/or the type of serial connection.

Please let me know if you want me to test you alternative solution with mpremote and MicroPython

